PR #76, opened 2024-12-21
update typo: Boost_LIBRARYDIR -> BOOST_LIBRARYDIR

kitaniboy:
There is a typo in the CMAKE environment variable; needs to fix "Boost_LIBRARYDIR -> BOOST_LIBRARYDIR".

Without the fix, the custom path for boost library would not work on linux cmake configuration.


CMakeLists.txt
```diff
@@ -103,7 +103,8 @@ if (WIN32)
     set(Boost_THREADAPI  win32)
 else()
     set(BOOST_INCLUDEDIR "/usr/local/include/boost")
-    set(Boost_LIBRARYDIR "/usr/local/lib/")
+    set(BOOST_LIBRARYDIR "/usr/local/lib")
+
```
luminousmining, line 107: Remove empty line please.

luminousmining:
Thank you!
Can you remove empty line please and I merge it.

CMakeLists.txt
luminousmining: https://cmake.org/cmake/help/latest/module/FindBoost.html
BOOST_LIBRARYDIR
Preferred library directory e.g. <prefix>/lib.


Maybe we shoudl remove `/lib` and keep only `set(BOOST_LIBRARYDIR "/usr/local")`

kitaniboy: The <prefix> is set by `BOOST_ROOT`. I think "/usr/local/lib" is the correct default value for `BOOST_LIBRARYDIR`.

kitaniboy:
Okay. The empty line is cleaned up now.

PR #79, opened 2024-12-23
support custom api port setting

kitaniboy:
This update introduces a new parameter `api_port=8080`, which replaces the fixed `8080` port of the `serverAPI` with the parameter.

https://github.com/luminousmining/miner/blob/79a4ad84ca73fca2948225dc40a102ba2e44f96c/sources/api/api.cpp#L42

The fixed api port `8080` perhaps is the default setting for HiveOS (I don't know well about HiveOS though). However, in general, the port `8080` is frequently occupied by other applications such as terminal nodes for linux systems. In this case, the system throws an error upon binding the api port, which then terminates the miner process.

### DIFF
--- a/sources/common/cli_pool.cpp
+++ b/sources/common/cli_pool.cpp
@@ -117,3 +117,12 @@ std::optional<std::string> common::Cli::getPassword() const
     }
     return std::nullopt;
 }
+
+uint32_t common::Cli::getApiPort() const
+{
+    if (true == contains("api_port"))
+    {
+        return params["api_port"].as<uint32_t>();
+    }
+    return 8080u;
+}
--- a/sources/common/cli_pool.cpp
+++ b/sources/common/cli_pool.cpp
@@ -118,6 +118,7 @@ std::optional<std::string> common::Cli::getPassword() const
     return std::nullopt;
 }
 
+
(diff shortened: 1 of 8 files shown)


sources/common/cli_pool.cpp
```diff
@@ -117,3 +117,12 @@ std::optional<std::string> common::Cli::getPassword() const
     }
     return std::nullopt;
 }
+
+uint32_t common::Cli::getApiPort() const
```
luminousmining, line 121: I forgot to indicate in the CODYNG_STYLE.md that it is necessary to put 2 empty lines between the functions.

sources/common/cli_pool.cpp
```diff
@@ -117,3 +117,12 @@ std::optional<std::string> common::Cli::getPassword() const
...
+{
+    if (true == contains("api_port"))
+    {
+        return params["api_port"].as<uint32_t>();
+    }
+    return 8080u;
+}
```
luminousmining, line 128: Add new line at end file please.

luminousmining:
Nice update !
Need 2 modifications about coding style and it's perfect !

kitaniboy:
Resolved the coding style issue. Thanks for the review.

PR #80, opened 2024-12-24
support socks5 proxy through a localhost server

kitaniboy:
This update introduces a new feature, connecting to a pool via socks5 server in localhost (e.g., [tor proxy](https://www.torproject.org/), [ssh dynamic port forwarding](https://man.openbsd.org/ssh#D), etc.).

The feature is enabled if the parameter `socks5` is `true`, where the socks5 port is specified by the parameter `socks_port` (default is `9050`, which is the default of the tor linux client).

Rented GPU nodes (e.g., [clore.ai](https://clore.ai/)) often blocks the mining pool addresses for some reasons. In this case, the socks5 proxy is a useful solution to bypass the network blocking.

For socks5 client implementation, I borrowed a boost::asio-based client in https://github.com/sehe/asio-socks45-client. Currently, only a socks5 server on localhost without id & password authentications is supported. This suffices the use cases when tor proxy or ssh dynamic port forwarding is used (nevertheless, I believe it should not be too difficult to update codes for a more general socks5 support).

### DIFF
--- a/sources/common/cli.cpp
+++ b/sources/common/cli.cpp
@@ -92,6 +92,19 @@ common::Cli::Cli()
             "Default value is false.\n"
             "--stale=<true|false>"
         )
+        (
+            "socks5",
+            value<bool>(),
+            "[OPTIONAL] Enable SOCKS5 proxy.\n"
+            "Default value is false.\n"
+            "--socks5=<true|false>."
+        )
+        (
+            "socks_port",
+            value<uint32_t>(),
+            "[OPTIONAL] Port of the SOCKS proxy.\n"
+            "--socks_port=9050"
+        )
 
         // Pool Custom
         (
--- a/sources/network/socks5.hpp
+++ b/sources/network/socks5.hpp
@@ -1,3 +1,5 @@
+// Borrowed from https://github.com/sehe/asio-socks45-client/blob/efeae8933348908768f1764f04fbe3839248bfa8/socks5.hpp
+
 #include <boost/asio.hpp>
 #include <boost/endian/arithmetic.hpp>
 
--- a/sources/network/socks5.hpp
+++ b/sources/network/socks5.hpp
@@ -1,0 +1,444 @@
+#include <boost/asio.hpp>
+#include <boost/endian/arithmetic.hpp>
+
+namespace socks5 { // threw in the kitchen sink for error codes
+#ifdef STANDALONE_ASIO
+    using std::error_category;
+    using std::error_code;
+    using std::error_condition;
+    using std::system_error;
+#else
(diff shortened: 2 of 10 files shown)


sources/common/cli.cpp
```diff
@@ -92,6 +92,19 @@ common::Cli::Cli()
...
         )
+        (
+            "socks5",
+            value<bool>(),
+            "[OPTIONAL] Enable SOCKS5 proxy.\n"
+            "Default value is false.\n"
+            "--socks5=<true|false>."
+        )
+        (
+            "socks_port",
```
luminousmining, line 103: Idea is good but multi parameters can be confuse for users.
Continue to use `--port` and keep `--socks5` remove `--socks_port`.

In the function `boost_endpoint    socksEndpoint{ {}, static_cast<boost::asio::ip::port_type>(config.mining.socksPort) };` replace `config.mining.socksPort` by `port` have the same effect.

The user must set `--port` and add `--socks5` to enable this feature.

sources/network/socks5.hpp
```diff
@@ -0,0 +1,444 @@
+#include <boost/asio.hpp>
```
luminousmining, line 1: Can you add credit for `sehe` about `https://github.com/sehe/asio-socks45-client` in REDAME.md please in the section `#Reference` and add new subsection `#References - Librairies`

sources/common/cli.cpp
luminousmining: we can set `port` a `9050` by default if `--socks5` is enable and user did not define `--port`,

kitaniboy: I think we have some confusion here between `port` (the port of a stratum server) and `socks_port` (socks5 server running in localhost). Both of the parameters are need for connecting to a stratum via socks5 proxy server.

Nevertheless, I agree that the current parameter definition could be confusion for users. How about we use `--proxy=HOST:PORT` (following [xmrig](https://github.com/xmrig/xmrig)) and enable socks5 when `proxy` is not null?

sources/network/socks5.hpp
kitaniboy: Added a recognition of the author (`sehe`)

sources/common/cli.cpp
luminousmining: Ho ! My bad ok thanks ! I confused it. I'm ok with this solution. Please add a comment in this parameters for user.
Thanks you !

kitaniboy: Thanks for the review. I have pushed more detailed comments and examples regarding socks5 parameters.

PR #133, opened 2026-06-05
fix: always authorize after mining.subscribe (NiceHash EthereumStratum)

yuzi-co:
## Problem
In `StratumEthash::onResponse`, the `ID_MINING_SUBSCRIBE` handler only called `miningAuthorize()` **inside** the extranonce-parse block. NiceHash's `EthereumStratum/1.0.0` returns the extranonce as `result[1]` of a nested array; any hiccup reading `result.at(1)` (wrong shape / short array → out-of-range) skipped authorize entirely.

## Impact
The worker stays unauthenticated, so the pool rejects **every** share. On NiceHash this surfaces as the misleading `Missing KYC/KYB` message. Reproduced live mining daggerhashimoto on NiceHash.

## Fix
Parse the subscribe reply defensively (require `result` to be an array of size ≥ 2 with a string at index 1) and **always** call `miningAuthorize()` regardless of extranonce parse outcome. Also handle a non-null `error` explicitly.

### DIFF
--- a/sources/stratum/ethash.cpp
+++ b/sources/stratum/ethash.cpp
@@ -20,9 +20,8 @@ void stratum::StratumEthash::onResponse(boost::json::object const& root)
         {
             if (false == root.contains("error") || true == root.at("error").is_null())
             {
-                if (true == root.contains("result") && true == root.at("result").is_array())
+                if (stratum::STRATUM_TYPE::ETHEREUM_V1 == stratumType)
                 {
-                    boost::json::array const& result{ root.at("result").as_array() };
                     // NiceHash (EthereumStratum/1.0.0) returns the extranonce as the
                     // second element of "result": [["mining.notify", id, proto], nonce].
                     // Parse it defensively: a missing/short/oddly-typed result must NOT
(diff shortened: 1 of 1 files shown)


sources/stratum/ethash.cpp
```diff
@@ -18,22 +18,29 @@ void stratum::StratumEthash::onResponse(boost::json::object const& root)
     {
         case stratum::Stratum::ID_MINING_SUBSCRIBE:
         {
-            if (false == root.contains("error") || true == root.at("error").is_null())
+            if (true == root.contains("error") && false == root.at("error").is_null())
```
luminousmining: The PR correctly inverts the error handling logic: instead of continuing if everything is fine, we explicitly stop if an error is detected via `if (true == root.contains("error") && false == root.at("error").is_null())`. This is a valid approach, but if we adopt this pattern here, it should be applied consistently across all functions of the different stratums to keep the codebase harmonious.

sources/stratum/ethash.cpp
```diff
@@ -18,22 +18,29 @@ void stratum::StratumEthash::onResponse(boost::json::object const& root)
...
+            {
+                logErr() << "Subscribe failed : " << root;
+                break;
+            }
+            // NiceHash (EthereumStratum/1.0.0) returns the extranonce as the
```
luminousmining: The comment would be better placed directly above `if (result.size() >= 2u && true == result.at(1).is_string())` rather than at the top of the block, so it sits as close as possible to the code it explains and makes it immediately clear why we specifically check that `.at(1)` is a string.

sources/stratum/ethash.cpp
yuzi-co: Reworked it to stay in the existing `if (false == root.contains("error") || true == root.at("error").is_null()) { ... } else { ... }` shape the other handlers already use, rather than introduce the early-return style -- so there's no new pattern to propagate across `autolykos_v2`/`progpow`/etc. The extranonce parse now nests in the success branch and `miningAuthorize()` always runs on a non-error reply. `ID_MINING_AUTHORIZE` is left unchanged.

sources/stratum/ethash.cpp
yuzi-co: Moved the comment directly above the `if (result.size() >= 2u && true == result.at(1).is_string())` check.

PR #136, opened 2026-06-05
fix(amd): build the whole DAG when the driver clamps the launch (RDNA4 ethash)

yuzi-co:
## Problem

`ethash_build_dag` generates the DAG with one work-item per page and a 2D launch of `(maxGroupSize, numberItem / maxGroupSize)`. On some drivers — observed on **AMD gfx1201 / RDNA4 (RX 9070 XT)** — the runtime silently executes far fewer work-groups than requested when a launch dimension is very large. Only the first ~2^24 (16,777,216) pages get written; the rest of a multi-GB DAG stays **zero**.

Hashing over the zeroed pages produces wrong results, so **every share is rejected (`Share above target`)**. Algorithms whose DAG fits under the limit (e.g. kawpow at low epochs, ~2.1 GB) are unaffected — which is why progpow worked while mainnet ethash epochs (5.77 GB, ~45M pages) did not.

## Fix

- **`sources/algo/ethash/opencl/ethash_dag.cl`** — grid-stride over every page so coverage no longer depends on how many work-items actually run.
- **`sources/resolver/amd/ethash.cpp` & `sources/resolver/amd/progpow.cpp`** — cap the launch Y dimension (32768) so the dispatch is complete and `get_global_size()` reflects the real launch; the kernel grid-strides to cover the rest.

The search kernel is unchanged. The change is a no-op for DAGs that already fit in one launch.

The **NVIDIA CUDA** build path is **not affected**: `ethashBuildDag` (`sources/algo/ethash/cuda/dag.cuh`) already builds the DAG in a host-side loop of small `<<<512, 128>>>` launches with a `startIndex` offset, so no launch dimension approaches the grid limit. Only the AMD path issued a single oversized launch.

## Verification (RX 9070 XT, OpenCL)

- `ResolverEthashAmdTest` known-answer `findNonce` (epoch 560) — **passes** (failed before the fix).
- Live NiceHash `daggerhashimoto` ethash — **accepted shares, ~31 MH/s** (was 100% rejected).
- `ResolverKawpowAmdTest.findNonce` / `ResolverProgpowQuaiAmdTest.findNonce` — still pass.

### DIFF
--- a/sources/algo/ethash/opencl/ethash_dag.cl
+++ b/sources/algo/ethash/opencl/ethash_dag.cl
@@ -33,12 +33,17 @@ void ethash_build_dag(
     uint const cache_number_item)
 {
     ////////////////////////////////////////////////////////////////////////////
-    uint const dag_index = get_global_id(0) + get_global_id(1) * GROUP_SIZE;
-    if (dag_index >= dag_number_item)
+    // Grid-stride over every DAG page. Some drivers (e.g. AMD on gfx1201/RDNA4)
+    // silently execute fewer work-groups than requested when a launch dimension is
+    // very large, leaving most of a big DAG (multi-GB ethash epochs) unwritten and
+    // zero. Striding by the real launch size makes coverage independent of how many
+    // work-items actually run. The host also caps the launch so it is fully
+    // dispatched (see resolver buildDAG).
+    uint const grid_stride = (uint)(get_global_size(0) * get_global_size(1));
+    for (uint dag_index = (uint)(get_global_id(0) + get_global_id(1) * get_global_size(0));
+         dag_index < dag_number_item;
+         dag_index += grid_stride)
     {
-        return;
-    }
-
     ////////////////////////////////////////////////////////////////////////////
     __attribute__((opencl_unroll_hint))
     for (uchar loop = 0; loop < 2; ++loop)
@@ -87,5 +92,6 @@ void ethash_build_dag(
             dag[gap_index + x] = item[x];
         }
 
+    }
     }
 }
--- a/sources/algo/ethash/opencl/ethash_dag.cl
+++ b/sources/algo/ethash/opencl/ethash_dag.cl
@@ -44,54 +44,58 @@ void ethash_build_dag(
          dag_index < dag_number_item;
          dag_index += grid_stride)
     {
-    ////////////////////////////////////////////////////////////////////////////
-    __attribute__((opencl_unroll_hint))
-    for (uchar loop = 0; loop < 2; ++loop)
-    {
         ////////////////////////////////////////////////////////////////////////
-        uint const cache_start_index = (dag_index * 2u) + loop;
-        uint const cache_index = (cache_start_index % cache_number_item) * 4u;
-
-        ////////////////////////////////////////////////////////////////////////
-        uint4 item[4];
         __attribute__((opencl_unroll_hint))
-        for (uchar i = 0; i < 4; ++i)
+        for (uchar loop = 0; loop < 2; ++loop)
         {
-            uint const index_tmp = cache_index + i;
-            item[i] = cache[index_tmp];
-        }
-        item[0].x ^= cache_start_index;
+            ////////////////////////////////////////////////////////////////////
+            uint const cache_start_index = (dag_index * 2u) + loop;
+            uint const cache_index = (cache_start_index % cache_number_item) * 4u;
 
-        ////////////////////////////////////////////////////////////////////////
-        keccak_f1600(item);
-
-        ////////////////////////////////////////////////////////////////////////
-        uint k = 0;
-        __attribute__((opencl_unroll_hint))
-        for (uchar i = 0; i < DAG_LOOP; ++i)
-        {
+            ////////////////////////////////////////////////////////////////////
+            uint4 item[4];
             __attribute__((opencl_unroll_hint))
-            for (uchar j = 0; j < 4u; ++j)
+            for (uchar i = 0; i < 4; ++i)
             {
-                build_item_mix(cache, item, cache_number_item, fnv1_u32(cache_start_index ^ k,        item[j & 3u].x));
-                build_item_mix(cache, item, cache_number_item, fnv1_u32(cache_start_index ^ (k + 1u), item[j & 3u].y));
-                build_item_mix(cache, item, cache_number_item, fnv1_u32(cache_start_index ^ (k + 2u), item[j & 3u].z));
-                build_item_mix(cache, item, cache_number_item, fnv1_u32(cache_start_index ^ (k + 3u), item[j & 3u].w));
-                k += 4u;
+                uint const index_tmp = cache_index + i;
+                item[i] = cache[index_tmp];
             }
+            item[0].x ^= cache_start_index;
+
+            ////////////////////////////////////////////////////////////////////
+            keccak_f1600(item);
+
+            ////////////////////////////////////////////////////////////////////
+            uint k = 0;
+            __attribute__((opencl_unroll_hint))
+            for (uchar i = 0; i < DAG_LOOP; ++i)
+            {
+                __attribute__((opencl_unroll_hint))
+                for (uchar j = 0; j < 4u; ++j)
+                {
+                    build_item_mix(cache, item, cache_number_item,
+                        fnv1_u32(cache_start_index ^ k, item[j & 3u].x));
+                    build_item_mix(cache, item, cache_number_item,
+                        fnv1_u32(cache_start_index ^ (k + 1u), item[j & 3u].y));
+                    build_item_mix(cache, item, cache_number_item,
+                        fnv1_u32(cache_start_index ^ (k + 2u), item[j & 3u].z));
+                    build_item_mix(cache, item, cache_number_item,
+                        fnv1_u32(cache_start_index ^ (k + 3u), item[j & 3u].w));
+                    k += 4u;
+                }
+            }
+
+            ////////////////////////////////////////////////////////////////////
+            keccak_f1600(item);
+
+            ////////////////////////////////////////////////////////////////////
+            uint const gap_index = (dag_index * 8u) + (loop * 4);
+            __attribute__((opencl_unroll_hint))
+            for (uint x = 0u; x < 4u; ++x)
+            {
+                dag[gap_index + x] = item[x];
+            }
+
         }
-
-        ////////////////////////////////////////////////////////////////////////
-        keccak_f1600(item);
-
-        ////////////////////////////////////////////////////////////////////////
-        uint const gap_index = (dag_index * 8u) + (loop * 4);
-        __attribute__((opencl_unroll_hint))
-        for (uint x = 0u; x < 4u; ++x)
-        {
-            dag[gap_index + x] = item[x];
-        }
-
-    }
     }
 }
--- a/sources/resolver/amd/ethash.cpp
+++ b/sources/resolver/amd/ethash.cpp
@@ -259,9 +259,18 @@ bool resolver::ResolverAmdEthash::buildDAG()
     OPENCL_ER(clKernel.setArg(4u, castU32(context.lightCache.numberItem)));
 
     ////////////////////////////////////////////////////////////////////////////
-    // Run kernel to build DAG
+    // Run kernel to build DAG.
+    // Cap the Y dimension so the whole launch is actually dispatched: some drivers
+    // (AMD on gfx1201/RDNA4) silently execute fewer work-groups than requested when
+    // a dimension is very large, which leaves most of a multi-GB DAG zero. The
+    // kernel grid-strides, so a capped launch still covers every page.
     uint32_t const maxGroupSize{ getMaxGroupSize() };
-    uint32_t const threadKernel{ castU32(context.dagCache.numberItem) / maxGroupSize };
+    uint32_t const maxThreadKernel{ 32768u };
+    uint32_t       threadKernel{ (castU32(context.dagCache.numberItem) + maxGroupSize - 1u) / maxGroupSize };
+    if (threadKernel > maxThreadKernel)
+    {
+        threadKernel = maxThreadKernel;
+    }
     OPENCL_ER(clQueue[currentIndexStream]->enqueueNDRangeKernel(
         clKernel,
         cl::NullRange,
--- a/sources/resolver/amd/ethash.cpp
+++ b/sources/resolver/amd/ethash.cpp
@@ -260,21 +260,28 @@ bool resolver::ResolverAmdEthash::buildDAG()
 
     ////////////////////////////////////////////////////////////////////////////
     // Run kernel to build DAG.
-    // Cap the Y dimension so the whole launch is actually dispatched: some drivers
-    // (AMD on gfx1201/RDNA4) silently execute fewer work-groups than requested when
-    // a dimension is very large, which leaves most of a multi-GB DAG zero. The
-    // kernel grid-strides, so a capped launch still covers every page.
-    uint32_t const maxGroupSize{ getMaxGroupSize() };
-    uint32_t const maxThreadKernel{ 32768u };
-    uint32_t       threadKernel{ (castU32(context.dagCache.numberItem) + maxGroupSize - 1u) / maxGroupSize };
-    if (threadKernel > maxThreadKernel)
+    // The build kernel grid-strides over the DAG (for (i = gid; i < numberItem;
+    // i += stride)), so coverage is complete for any launch of at least one
+    // work-group. We cap the work-group count (the NDRange Y dimension) because
+    // some drivers (AMD on gfx1201/RDNA4) silently dispatch fewer work-groups than
+    // requested when a dimension is very large, which leaves most of a multi-GB
+    // DAG zero. MAX_BUILD_GROUPS is therefore a ceiling, not a tuning constant: it
+    // only has to stay below that silent-clamp threshold while still saturating the
+    // CUs. 32768 groups (~8M work-items) is already orders of magnitude above the
+    // simultaneous occupancy of any current GPU, so it saturates every model; the
+    // build is DRAM-bandwidth bound, so a larger or per-device value would not
+    // change throughput.
+    constexpr uint32_t MAX_BUILD_GROUPS{ 32768u };
+    uint32_t const     maxGroupSize{ getMaxGroupSize() };
+    uint32_t           buildGroups{ (castU32(context.dagCache.numberItem) + maxGroupSize - 1u) / maxGroupSize };
+    if (buildGroups > MAX_BUILD_GROUPS)
     {
-        threadKernel = maxThreadKernel;
+        buildGroups = MAX_BUILD_GROUPS;
     }
     OPENCL_ER(clQueue[currentIndexStream]->enqueueNDRangeKernel(
         clKernel,
         cl::NullRange,
-        cl::NDRange(maxGroupSize, threadKernel, 1),
+        cl::NDRange(maxGroupSize, buildGroups, 1),
         cl::NDRange(maxGroupSize, 1, 1)));
     OPENCL_ER(clQueue[currentIndexStream]->finish());
 
--- a/sources/resolver/amd/tests/firopow.cpp
+++ b/sources/resolver/amd/tests/firopow.cpp
@@ -84,7 +84,7 @@ TEST_F(ResolverFiropowAmdTest, findNonce)
 TEST_F(ResolverFiropowAmdTest, notFindNonce)
 {
     initializeJob(0x85f22c9b3cd2f123);
-    jobInfo.boundaryU64 = 1ull;  // unsatisfiable boundary -> no share can clear it
+    jobInfo.boundaryU64 = 1ull; // unsatisfiable boundary -> no share can clear it
 
     ASSERT_TRUE(resolver.updateMemory(jobInfo));
     ASSERT_TRUE(resolver.updateConstants(jobInfo));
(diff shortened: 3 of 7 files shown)


sources/resolver/amd/tests/firopow.cpp
```diff
@@ -38,27 +38,30 @@ struct ResolverFiropowAmdTest : public testing::Test
...
+    // block 1 / epoch 0. Low epoch keeps the DAG small and the hash correct
+    // (see the DISABLED epoch-419 test for the known large-DAG defect).
     void initializeJob(uint64_t const nonce)
     {
         jobInfo.nonce = nonce;
-        jobInfo.blockNumber = 544860ull;
-        jobInfo.headerHash = algo::toHash256("23bb6340c5ffcf05c3e2b86503a636b805f7ca6c93d50315971b26b72f461790");
-        jobInfo.seedHash = algo::toHash256("ac88bf0324754ae04cffe412accbbd72e534acd928bdcbe95239f660667ae26d");
-        jobInfo.boundary = algo::toHash256("00000003fffc0000000000000000000000000000000000000000000000000000");
+        jobInfo.blockNumber = 1ull;
```
luminousmining, line 54: These data come from a real captured job.
In the unit tests I prefer to have a test that matches reality as closely as possible.
In your example the block number is too low.

sources/resolver/amd/tests/firopow.cpp
```diff
@@ -38,27 +38,30 @@ struct ResolverFiropowAmdTest : public testing::Test
...
+        jobInfo.headerHash = algo::toHash256("2d794e900dcad779e658de9078d9a88eee87d75f7b09a8fdd270d3a8e76650c7");
+        jobInfo.boundary = algo::toHash256("0001869e7a058e2aaf266cd2f166fb85c98d651e60eadbbe72bb0a36f8802805");
         jobInfo.boundaryU64 = algo::toUINT64(jobInfo.boundary);
-        jobInfo.epoch =
-            algo::ethash::ContextGenerator::instance().findEpoch(jobInfo.seedHash, algo::firopow::EPOCH_LENGTH);
+        jobInfo.epoch = 0;
```
luminousmining: forcing `jobInfo.epoch = 0;` removes the implicit epoch calculation test.

sources/resolver/amd/tests/firopow.cpp
```diff
@@ -38,27 +38,30 @@ struct ResolverFiropowAmdTest : public testing::Test
         properties.clQueue = nullptr;
     }
 
+    // Authoritative FiroPoW vector (firoorg/firo firopow_test_vectors.hpp),
```
luminousmining, line 47: I'm ok to keep test with block at 1 and epoch at 0 but we need a test matches reality as closely as possible.

sources/resolver/amd/ethash.cpp
```diff
@@ -184,9 +184,18 @@ bool resolver::ResolverAmdEthash::buildDAG()
...
     ////////////////////////////////////////////////////////////////////////////
-    // Run kernel to build DAG
+    // Run kernel to build DAG.
+    // Cap the Y dimension so the whole launch is actually dispatched: some drivers
+    // (AMD on gfx1201/RDNA4) silently execute fewer work-groups than requested when
+    // a dimension is very large, which leaves most of a multi-GB DAG zero. The
+    // kernel grid-strides, so a capped launch still covers every page.
     uint32_t const maxGroupSize{ getMaxGroupSize() };
-    uint32_t const threadKernel{ castU32(context.dagCache.numberItem) / maxGroupSize };
+    uint32_t const maxThreadKernel{ 32768u };
```
luminousmining: Why 32768? Shouldn't this value be dynamic based on the GPU model being used?

sources/resolver/amd/tests/firopow.cpp
```diff
@@ -67,18 +70,114 @@ TEST_F(ResolverFiropowAmdTest, findNonce)
...
 
 TEST_F(ResolverFiropowAmdTest, notFindNonce)
 {
-    initializeJob(0x000100000704757f);
+    initializeJob(0x85f22c9b3cd2f123);
+    jobInfo.boundaryU64 = 1ull;  // unsatisfiable boundary -> no share can clear it
 
     ASSERT_TRUE(resolver.updateMemory(jobInfo));
     ASSERT_TRUE(resolver.updateConstants(jobInfo));
+    resolver.setThreads(1u);
```
luminousmining, line 91: I understand that it is faster to execute one thread instead of several.
But this creates a "silent" problem — the kernel works in collaboration with multiple lanes (threads) `LANE_PARALLEL` at the same time.
It is better to keep the base GRID so that the required number of lanes is present.

sources/resolver/amd/tests/firopow.cpp
```diff
@@ -67,18 +70,114 @@ TEST_F(ResolverFiropowAmdTest, findNonce)
...
     EXPECT_TRUE(stratum.paramSubmit.empty());
 }
+
+
+// Verifies the *submitted mixhash* against authoritative FiroPoW test vectors
+// (firoorg/firo: src/crypto/progpow/firopow_test_vectors.hpp). The existing tests
+// only check that the nonce is found; a wrong-but-self-consistent digest would
+// still pass them yet be rejected live as "Invalid Firo Mixhash". These pin the
+// digest itself, across epochs (epoch 0 does not exercise DAG growth/seed).
+namespace
```
luminousmining: Can you encapsulate it in a proper namespace `resolver` + `test` ?

sources/resolver/amd/tests/firopow.cpp
```diff
@@ -67,18 +70,114 @@ TEST_F(ResolverFiropowAmdTest, findNonce)
...
+            }
+        }
+        EXPECT_TRUE(found) << "vector nonce was not among submitted shares";
+    }
+}
+
+
+TEST_F(ResolverFiropowAmdTest, submitMixMatchesFiroVector)  // block 1, epoch 0
+{
+    checkFiroVector(*this, 0x85f22c9b3cd2f123ull, 1ull, 0,
```
luminousmining: Can you add a line break for each parameter? This should be applied to every call to `checkFiroVector`.

sources/resolver/amd/tests/firopow.cpp
```diff
@@ -67,18 +70,114 @@ TEST_F(ResolverFiropowAmdTest, findNonce)
...
+        t.jobInfo.boundary    = algo::toHash256(boundaryHex);
+        t.jobInfo.boundaryU64 = algo::toUINT64(t.jobInfo.boundary);
+        t.jobInfo.epoch       = static_cast<int32_t>(epoch);
+        t.jobInfo.period      = t.jobInfo.blockNumber / algo::firopow::MAX_PERIOD;
+
+        ASSERT_TRUE(t.resolver.updateMemory(t.jobInfo));
+        ASSERT_TRUE(t.resolver.updateConstants(t.jobInfo));
+        // One workgroup (256 nonces) starting at the vector nonce: the vector nonce
+        // is computed and essentially nothing else clears the boundary.
+        t.resolver.setThreads(1u);
```
luminousmining, line 129: Same remark than `LANE_PARALLEL`
Theoretically the resolver should find no "nonce" answer, as a single thread cannot function correctly.

sources/resolver/amd/ethash.cpp
yuzi-co: It's a ceiling, not a tuning constant, and it's deliberately GPU-independent.

The build kernel grid-strides over the DAG (`for (i = gid; i < numberItem; i += stride)`), so coverage is complete for *any* launch size >= 1 group -- correctness doesn't depend on the value. The cap exists only to keep the Y dimension small enough that the gfx1201/RDNA4 driver doesn't silently drop work-groups (the bug this PR fixes): an uncapped launch requests `ceil(numberItem / groupSize)` groups, which for a multi-GB DAG is large enough to trip the clamp.

So there are only two constraints: (1) below the driver's silent-clamp threshold, (2) enough groups to saturate the CUs. 32768 groups (x `groupSize` ~= 8M work-items) is already orders of magnitude above the simultaneous occupancy of any current GPU, so it saturates every model while staying well under the clamp. Deriving it per-device would add an occupancy query without changing correctness or measurably changing build throughput -- the build is DRAM-bandwidth bound, not launch-occupancy bound.

Happy to make it explicit though: I can pull it out as a named `constexpr maxBuildGroups` with this rationale in the comment, or derive a ceiling from CU count x waves-per-CU if you'd prefer it look dynamic. Which do you want?

sources/resolver/amd/tests/firopow.cpp
yuzi-co: I think this is a granularity mix-up on `setThreads` -- it sets the count of *work-groups* (the Y dimension), not work-items. The search launch is `NDRange(blocks, threads, 1)` global with `NDRange(blocks, 1, 1)` local, and `blocks == getMaxGroupSize()` (= `GROUP_SIZE`, 256 here). So `setThreads(1u)` still dispatches one **full** work-group of 256 work-items = 16 `LANES` x 16 lane-groups -- every lane is present and cooperates exactly as in production. It just launches a single group instead of thousands.

That one group covers `GROUP_SIZE/LANES` = 16 consecutive nonces from `jobInfo.nonce`, which is why the vector nonce is found deterministically and essentially nothing else clears the boundary -- that determinism is the point of using one group. A genuinely single-*threaded* launch would indeed find nothing, but that's not what this does. I'll fix the misleading comment (it says "256 nonces" -- should be "16 nonces / 256 lanes") so this isn't confusing on the next read.

sources/resolver/amd/tests/firopow.cpp
yuzi-co: Same as the checkFiroVector thread below: setThreads(1u) is one full 256-work-item group (16 lanes present), not a single thread -- so the kernel runs correctly. notFindNonce just sets an unsatisfiable boundaryU64 (1), so that full group runs and submits nothing, which is what's asserted.

sources/algo/ethash/opencl/ethash_dag.cl
```diff
@@ -88,4 +93,5 @@ void ethash_build_dag(
         }
 
     }
+    }
```
luminousmining, line 100: fix indent

sources/resolver/amd/tests/firopow.cpp
yuzi-co: The four vectors here are the authoritative `firoorg/firo` `firopow_test_vectors.hpp` set, so they're pinned to upstream's own reference digests rather than a single captured job -- that's deliberate, it lets the test catch a wrong-but-self-consistent mixhash against Firo's canonical values.

For the "matches reality" angle, `submitMixMatchesFiroVectorEpoch419` (block 545860 -> epoch 419, ~5 GB DAG) is exactly the regime real Firo blocks mine in today, and it's the direct regression for this PR: without the grid-stride fix the multi-GB DAG tail stays zero, so that case fails while the low-block ones pass. So the realistic case is already covered by that test; the low blocks (1 / 1300 / 13000) stay as the small-DAG reference points.

Does the epoch-419 vector satisfy the "close to reality" concern, or would you still like a literal captured-from-pool job added alongside it?

sources/resolver/amd/tests/firopow.cpp
yuzi-co: Kept block 1 as you're ok with -- its realistic counterpart is the epoch-419 / block 545860 test below (the regime real Firo blocks mine in, ~5 GB DAG), which is also the direct regression for this fix.

sources/resolver/amd/tests/firopow.cpp
yuzi-co: Fixed -- the epoch is no longer hand-fed. It's now derived as `blockNumber / algo::firopow::EPOCH_LENGTH` (the same mapping production uses in `StratumProgPOW::deriveEpoch`), in both `initializeJob` and the `checkFiroVector` helper, and I dropped the explicit `epoch` parameter entirely. block 1 -> 0, 1300 -> 1, 13000 -> 10, 545860 -> 419 all fall out of the division (EPOCH_LENGTH = 1300).

sources/resolver/amd/tests/firopow.cpp
yuzi-co: Done -- the helper now lives in `namespace resolver::test`, with a `using resolver::test::checkFiroVector;` so the TEST_F bodies still call it unqualified.

sources/resolver/amd/tests/firopow.cpp
yuzi-co: Done -- every `checkFiroVector(...)` call now has one argument per line.

sources/resolver/amd/tests/firopow.cpp
luminousmining: Ok thanks, so I confused setThreads and setBlock for AMD resolver...
Threads should define work-item and block should define group.
I will fixe it later, to fix it I have been doing the tests very thoroughly.

sources/algo/ethash/opencl/ethash_dag.cl
yuzi-co: Fixed in `3a64b0c` -- the grid-stride loop wrapped the build body without re-indenting it, so the inner per-page loop sat at the same level as the grid-stride loop. Indented the body one level, re-anchored the section dividers back to column 80, and wrapped the four `build_item_mix` calls to stay within 120 cols. Whitespace only -- kernel tokens unchanged; re-verified ResolverEthashAmdTest + the FiroPoW vector suite (incl. epoch 419) still pass on gfx1201.

sources/resolver/amd/tests/firopow.cpp
yuzi-co: Thanks for confirming. So the test is correct as-is (one full work-group = all 16 lanes); no change needed here. Renaming the resolver's setThreads/setBlocks so the names match the semantics (threads = work-items, blocks = group) is a good separate cleanup -- happy to leave that to your follow-up.

sources/resolver/amd/ethash.cpp
yuzi-co: @
Went with the explicit named-constant option (pushed in `6bd3e0b`). The literal is now a `constexpr MAX_BUILD_GROUPS` with the rationale in the comment, and I renamed the misleading `maxThreadKernel`/`threadKernel` locals to `buildGroups` since the value is the work-group count (NDRange Y dimension), not threads — which also resolves the work-item/work-group mix-up from the LANE_PARALLEL threads below.

I kept it `UPPER_CASE` rather than the `maxBuildGroups` I quoted earlier because `.clang-tidy` enforces `readability-identifier-naming.ConstexprVariableCase: UPPER_CASE`.

Applied to both `ethash.cpp` and `progpow.cpp`, which share the identical clamp from this fix. Naming/comment only — no behavior change; I left it a fixed ceiling rather than a per-device query for the reasons above (grid-stride coverage is launch-size-independent and the build is DRAM-bandwidth bound).
@

sources/resolver/amd/tests/firopow.cpp
```diff
@@ -38,27 +38,31 @@ struct ResolverFiropowAmdTest : public testing::Test
...
-        jobInfo.headerHash = algo::toHash256("23bb6340c5ffcf05c3e2b86503a636b805f7ca6c93d50315971b26b72f461790");
-        jobInfo.seedHash = algo::toHash256("ac88bf0324754ae04cffe412accbbd72e534acd928bdcbe95239f660667ae26d");
-        jobInfo.boundary = algo::toHash256("00000003fffc0000000000000000000000000000000000000000000000000000");
+        jobInfo.blockNumber = 1ull;
+        jobInfo.headerHash = algo::toHash256("2d794e900dcad779e658de9078d9a88eee87d75f7b09a8fdd270d3a8e76650c7");
+        jobInfo.boundary = algo::toHash256("0001869e7a058e2aaf266cd2f166fb85c98d651e60eadbbe72bb0a36f8802805");
         jobInfo.boundaryU64 = algo::toUINT64(jobInfo.boundary);
-        jobInfo.epoch =
-            algo::ethash::ContextGenerator::instance().findEpoch(jobInfo.seedHash, algo::firopow::EPOCH_LENGTH);
+        jobInfo.epoch = static_cast<int32_t>(jobInfo.blockNumber / algo::firopow::EPOCH_LENGTH);
```
luminousmining: cast32()

sources/resolver/amd/tests/firopow.cpp
```diff
@@ -67,18 +71,130 @@ TEST_F(ResolverFiropowAmdTest, findNonce)
...
+        char const* const       expectedMix)
+    {
+        t.jobInfo.nonce       = nonce;
+        t.jobInfo.blockNumber = blockNumber;
+        t.jobInfo.headerHash  = algo::toHash256(headerHashHex);
+        t.jobInfo.boundary    = algo::toHash256(boundaryHex);
+        t.jobInfo.boundaryU64 = algo::toUINT64(t.jobInfo.boundary);
+        // Derive the epoch from the block number, as production does
+        // (StratumProgPOW::deriveEpoch), instead of hand-feeding it.
+        t.jobInfo.epoch       = static_cast<int32_t>(blockNumber / algo::firopow::EPOCH_LENGTH);
```
luminousmining: cast32()

sources/resolver/amd/tests/firopow.cpp
luminousmining: Thanks I will run a new session on Firo to take more informations.

sources/resolver/amd/tests/firopow.cpp
yuzi-co: Done in `b130c4f` — switched to the `cast32()` helper. Re-verified on the RX 9070 XT (cross-built `unit_test`, `GPU=amd`): `ResolverFiropowAmdTest.*` 6/6 pass, including the epoch-419 (~5 GB DAG) regression case.

sources/resolver/amd/tests/firopow.cpp
yuzi-co: Done in `b130c4f` — `cast32()` here too.

sources/resolver/amd/tests/firopow.cpp
```diff
@@ -73,18 +77,130 @@ TEST_F(ResolverFiropowAmdTest, findNonce)
...
+                found = true;
+                EXPECT_EQ(std::string(expectedMix), std::string(params[2].as_string().c_str()));
+            }
+        }
+        EXPECT_TRUE(found) << "vector nonce was not among submitted shares";
+    }
+}
+
+
+using resolver::test::checkFiroVector;
```
luminousmining: using only in function scope or use complet path `resolver::test::checkFiroVector::checkFiroVector` on each test

sources/stratum/tests/ethash_submit_nonce.cpp
```diff
@@ -0,0 +1,143 @@
...
+    ~StratumEthashSubmitNonceTest() override = default;
+};
+
+
+////////////////////////////////////////////////////////////////////////////////
+// setExtraNonce() places the pool extranonce in the HIGH bytes of the search
+// start nonce and records its hex width. This is what the resolver hashes, so it
+// must carry the extranonce -- otherwise the pool recomputes a different hash.
+////////////////////////////////////////////////////////////////////////////////
+
```
luminousmining: No blank line under function and comments

sources/resolver/amd/tests/firopow.cpp
yuzi-co: Done in `63d338d` — dropped the file-scope `using` and now call `resolver::test::checkFiroVector` with the full path in each of the four `TEST_F` bodies.

sources/stratum/tests/ethash_submit_nonce.cpp
yuzi-co: Done in `63d338d` — removed the blank line between the comment divider and the `TEST_F`. Applied to all four test dividers in the file for consistency, not just this one.

sources/stratum/tests/ethash_submit_nonce.cpp
```diff
@@ -0,0 +1,139 @@
...
+// convention so it is not "fixed" back into a rejection. The worked values below
+// are taken verbatim from the spec's own example (section III/IV):
+//
+//   extranonce = "a2eea0"  (3 bytes)
+//   minernonce = "cfae7df760"  (5 bytes, submitted as-is, no 0x)
+//   full nonce = 0xa2eea0cfae7df760
+////////////////////////////////////////////////////////////////////////////////
+
+
+namespace
```
luminousmining: no namespace empty

sources/stratum/tests/ethash_submit_nonce.cpp
```diff
@@ -0,0 +1,139 @@
...
+        nonceHexa << std::hex << nonce;
+        return nonceHexa.str().substr(extraNonceSize);
+    }
+}
+
+
+struct StratumEthashSubmitNonceTest : public testing::Test
+{
+    StratumEthashSubmitNonceTest() = default;
+    ~StratumEthashSubmitNonceTest() override = default;
```
luminousmining: override ?

sources/stratum/tests/ethash_submit_nonce.cpp
yuzi-co: Done in `2954150` — removed the unnamed namespace. The `SPEC_*` constants stay `constexpr` (already internal linkage) and `toMinerNonce` is now a `static` free function at file scope.

sources/stratum/tests/ethash_submit_nonce.cpp
yuzi-co: Done in `2954150` — dropped the `override` on the defaulted destructor, so the fixture now matches the repo convention (`~Foo() = default;`, as in the other test fixtures). Also rebased the branch onto current `main`.

PR #140, opened 2026-06-07
fix(stratum): derive progpow DAG epoch from seedHash, not block height

yuzi-co:
## Problem

FiroPoW pools (WoolyPooly, RPlant) rejected **every** share with `Invalid Firo Mixhash`.

`StratumProgPOW::onMiningNotify` computes the DAG epoch as `blockNumber / EPOCH_LENGTH`. Firo changed its FiroPoW epoch length between early and current blocks, so a fixed constant derives the wrong epoch for live blocks:

- block `1319805` → `1015` via `/1300`
- the real epoch encoded by the network `seedHash` is `650`

Wrong epoch → wrong DAG → wrong mix → 100% rejected.

## Fix

Derive the epoch from the `seedHash` via `ContextGenerator::findEpoch()` first — this is authoritative regardless of any epoch-length change, and is exactly what `StratumEthash` already does. Fall back to `blockNumber / EPOCH_LENGTH` only when the seed isn't recognized.

```cpp
int32_t epoch{ ContextGenerator::instance().findEpoch(jobInfo.seedHash, maxEthashEpoch) };
if (-1 == epoch && jobInfo.blockNumber > 0ull)
{
    epoch = cast32(jobInfo.blockNumber / castU64(maxEpochLength));
}
```

## Verification

- Live on WoolyPooly: FiroPoW now **accepted** (was 100% rejected); kawpow still accepted via the shared V1 path (no regression).
- Adds a hermetic regression test `EthashTest.firoEpochFromSeedHash` pinning a real Firo `seedHash` to epoch 650. The seed for a given epoch is fixed forever, so the assertion is stable as the chain advances.

Note: unit tests aren't built by the current CI (`BUILD_EXE_UNIT_TEST=OFF`); the test is for local runs / regression documentation.

Related but independent: the DAG-build coverage fix in #136 addresses a different large-DAG defect.

### DIFF
--- a/sources/stratum/progpow.cpp
+++ b/sources/stratum/progpow.cpp
@@ -323,7 +323,9 @@ void stratum::StratumProgPOW::onMiningNotify(boost::json::object const& root)
             // derives the wrong epoch (wrong DAG -> "Invalid Mixhash" reject) for live
             // Firo. Prefer the seed hash (as the ethash path already does); fall back
             // to blockNumber/EPOCH_LENGTH only if the seed isn't recognized.
-            int32_t epoch{ algo::ethash::ContextGenerator::instance().findEpoch(jobInfo.seedHash, maxEthashEpoch) };
+            // maxEpochSearch caps the seed search at a max epoch *number* (not a
+            // per-epoch length), so it must stay generous as the chain's epoch climbs.
+            int32_t epoch{ algo::ethash::ContextGenerator::instance().findEpoch(jobInfo.seedHash, maxEpochSearch) };
             if (-1 == epoch && jobInfo.blockNumber > 0ull)
             {
                 epoch = cast32(jobInfo.blockNumber / castU64(maxEpochLength));
@@ -393,7 +395,7 @@ void stratum::StratumProgPOW::onMiningNotify(boost::json::object const& root)
             }
             else
             {
-                epoch = algo::ethash::ContextGenerator::instance().findEpoch(jobInfo.seedHash, maxEthashEpoch);
+                epoch = algo::ethash::ContextGenerator::instance().findEpoch(jobInfo.seedHash, maxEpochSearch);
             }
             if (-1 != epoch)
             {
--- a/sources/stratum/progpow.cpp
+++ b/sources/stratum/progpow.cpp
@@ -317,14 +317,16 @@ void stratum::StratumProgPOW::onMiningNotify(boost::json::object const& root)
             }
 
             ////////////////////////////////////////////////////////////////////
-            int32_t epoch{ 0 };
-            if (jobInfo.blockNumber > 0ull)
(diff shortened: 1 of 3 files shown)


sources/stratum/progpow.cpp
```diff
@@ -317,15 +317,17 @@ void stratum::StratumProgPOW::onMiningNotify(boost::json::object const& root)
...
             ////////////////////////////////////////////////////////////////////
-            int32_t epoch{ 0 };
-            if (jobInfo.blockNumber > 0ull)
+            // The DAG epoch is whatever the network seedHash encodes — authoritative
+            // even when a chain changes its epoch length. Firo's FiroPoW epoch length
+            // differs between early and current blocks, so blockNumber/EPOCH_LENGTH
+            // derives the wrong epoch (wrong DAG -> "Invalid Mixhash" reject) for live
+            // Firo. Prefer the seed hash (as the ethash path already does); fall back
+            // to blockNumber/EPOCH_LENGTH only if the seed isn't recognized.
+            int32_t epoch{ algo::ethash::ContextGenerator::instance().findEpoch(jobInfo.seedHash, maxEthashEpoch) };
```
luminousmining: firopow test :
```cpp
        jobInfo.epoch =
            algo::ethash::ContextGenerator::instance().findEpoch(jobInfo.seedHash, algo::firopow::EPOCH_LENGTH);
```

Maybe we need change only `algo::ethash::ContextGenerator::instance().findEpoch(jobInfo.seedHash, maxEthashEpoch)` by `algo::ethash::ContextGenerator::instance().findEpoch(jobInfo.seedHash, maxEpochLength)` ?

Can you look it please

yuzi-co: The second argument to `findEpoch(seedHash, maxEpoch)` is the upper bound on the epoch *number* the seed search scans to — not a per-epoch length. `maxEpochLength` is a different quantity: the blocks-per-epoch divisor (`firopow::EPOCH_LENGTH = 1300`), used only in the fallback (`blockNumber / maxEpochLength`).

Swapping them would cap the FiroPoW seed search at epoch 1300. It works today (real epoch is 650), but FiroPoW's epoch number keeps climbing as the chain grows — once it passes 1300 the search would stop early and fall back to the wrong `blockNumber / EPOCH_LENGTH` again, the exact bug this fixes. The bound never changes *which* epoch matches (seeds are unique per epoch), only how far the search may scan.

You're right that the name was the confusing part though — `maxEthashEpoch` implied it was ethash-specific. I've renamed it to `maxEpochSearch` with clarifying comments (37ab9b5), keeping the value (`ethash::EPOCH_LENGTH = 30000`) since it's the same generous ceiling the ethash path already uses and it covers every realistic progpow/FiroPoW epoch.

PR #141, opened 2026-06-07
fix: portability of chrono.hpp / bitwise.hpp on non-AMD and non-MSVC builds

yuzi-co:
Two small, independent header-portability fixes.

### 1. `common/chrono.hpp` — missing `<string>` / `<cstdint>`

`std::string` and the fixed-width integer types were only available transitively via `<CL/opencl.hpp>`, which `cast.hpp` pulls in **solely** under `AMD_ENABLE`. NVIDIA / CPU builds therefore failed with:

```
no type named 'string' in namespace 'std'
```

Adds the two standard includes directly so the header is self-contained on every build configuration.

### 2. `algo/bitwise.hpp` — `<endian.h>` on non-Linux GNU compilers

The include was gated on `__GNUC__`, which pulls in `<endian.h>` for *any* GCC-compatible compiler. That header is glibc-specific — clang/MinGW on Windows defines `__GNUC__` but has no `<endian.h>`, so the include failed.

- `<stdlib.h>` path → `_WIN32` (covers MSVC **and** MinGW; MSVC also defines `_WIN32`, so its behaviour is unchanged)
- `<endian.h>` path → `__linux__`

Both are compile-only fixes (no behavioural change on the existing AMD/MSVC/Linux paths). No unit test is applicable.

### DIFF
--- a/sources/common/chrono.hpp
+++ b/sources/common/chrono.hpp
@@ -2,6 +2,8 @@
 
 
 #include <chrono>
+#include <cstdint>
+#include <string>
 
 
 namespace common
--- a/sources/algo/bitwise.hpp
+++ b/sources/algo/bitwise.hpp
@@ -2,9 +2,11 @@
 
 #if !defined(__LIB_CUDA)
 
-#if defined(_MSC_VER)
+#if defined(_WIN32)
+// Windows (MSVC and MinGW/clang): no <endian.h>. _byteswap_* live in <stdlib.h>;
+// byte-order macros are provided by the compiler or defined below.
 #include <stdlib.h>
-#elif defined(__linux__) || defined(__GNUC__)
+#elif defined(__linux__)
 #include <endian.h>
 #endif
 


sources/common/chrono.hpp
```diff
@@ -2,6 +2,8 @@
 
 
 #include <chrono>
+#include <cstdint>
```
luminousmining, line 5: Why include this files ?

yuzi-co: `chrono.hpp` uses both directly — it was just getting them transitively before:
- `<string>` — `ChronoGuard` takes `std::string const& text` and stores a `std::string text{}` member.
- `<cstdint>` — the `uint32_t` constants (`SEC_TO_MS`…), the `enum class CHRONO_UNIT : uint8_t`, and `uint64_t elapsed(...)`.

On AMD builds these arrived transitively via `<CL/opencl.hpp>` (pulled by `cast.hpp` only under `AMD_ENABLE`), so the header compiled. On NVIDIA/CPU builds that include isn't present and the header fails with `no type named 'string' in namespace 'std'`. The two includes just make the header self-contained on every configuration.

PR #145, opened 2026-06-07
fix(amd): GPU telemetry via ADL2 PMLog on modern GPUs (RDNA/Vega+)

yuzi-co:
## Problem

The AMD profiler reads GPU telemetry through `ADL_Overdrive5_CurrentActivity_Get` — the legacy Overdrive 5 API, which is unsupported on every AMD GPU since GCN 1.2 (Vega, all of RDNA 1–4). On those cards the call fails, the log fills with `[getCurrentActivity] ADL cannot get activity`, and the dashboard's clock / memory-clock / utilization read 0 (power was never read on the AMD path at all).

## Change

Reworks `profiler::Amd` into a backend-aware telemetry reader:

- **Primary — ADL2 PMLog:** an ADL2 context + `ADL2_New_QueryPMLogData_Get`, reading `ADL_PMLOG_*` sensors. Works on RDNA/Vega and newer. Reports core/memory clock (MHz), GPU utilization (%) and board power (W).
- **Fallback — Overdrive 5:** retained for pre-GCN 1.2 cards.
- **PCI-bus → ADL adapter-index map:** the ADL adapter index is neither the OpenCL device id nor the PCI bus, so adapters are enumerated and matched by bus number (which the device layer already tracks) — correct telemetry on multi-GPU rigs.
- **Heap-corruption fix:** the ADL malloc callback returned a fixed 1-byte block regardless of the requested size; it now honours the size.

The device stats loop consumes a small `Telemetry` struct and now surfaces AMD power, populating the previously-dead **Power / H/W / $/H** columns. This is the ADL-native path; an ADLX backend can later slot in behind the same `getTelemetry()` interface.

## Compatibility

Requires a PMLog-capable Adrenalin driver (standard on RDNA/Vega). Cards/drivers without PMLog fall back to Overdrive 5; if neither is available the profiler degrades gracefully to zeros (no crash, single warning).

## Testing

- Built for Windows (llvm-mingw cross) and Linux (clang-15) via the project's vcpkg toolchain; both link cleanly.
- Existing unit tests pass (non-GPU suites: 30/30).
- Live-verified on an **RX 9070 XT (RDNA4 / gfx1201)**: clock / utilization / power now display correctly and the ADL activity errors are gone.

### DIFF
--- a/sources/profiler/amd.cpp
+++ b/sources/profiler/amd.cpp
@@ -39,6 +39,49 @@ void* profiler::Amd::loadFunction(char const* name, bool const required)
 }
 
 
+////////////////////////////////////////////////////////////////////////////////
+/// Resolve the ADL2 (modern, context-based) entry points. Returns true only when
+/// the full set required to drive the PMLog path is present, so the caller can
+/// commit to Backend::PMLOG without re-checking individual pointers afterwards.
+bool profiler::Amd::loadAdl2Symbols()
+{
+    adl2MainControlCreate =
+        reinterpret_cast<ADL2_MAIN_CONTROL_CREATE>(loadFunction("ADL2_Main_Control_Create", false));
+    adl2MainControlDestroy =
+        reinterpret_cast<ADL2_MAIN_CONTROL_DESTROY>(loadFunction("ADL2_Main_Control_Destroy", false));
+    adl2AdapterNumberOfAdaptersGet =
+        reinterpret_cast<ADL2_ADAPTER_NUMBEROFADAPTERS_GET>(loadFunction("ADL2_Adapter_NumberOfAdapters_Get", false));
+    adl2AdapterAdapterInfoGet =
+        reinterpret_cast<ADL2_ADAPTER_ADAPTERINFO_GET>(loadFunction("ADL2_Adapter_AdapterInfo_Get", false));
+    adl2NewQueryPMLogDataGet =
+        reinterpret_cast<ADL2_PMLOG_QUERY_GET>(loadFunction("ADL2_New_QueryPMLogData_Get", false));
+
+    return nullptr != adl2MainControlCreate && nullptr != adl2NewQueryPMLogDataGet
+           && nullptr != adl2AdapterNumberOfAdaptersGet && nullptr != adl2AdapterAdapterInfoGet;
+}
+
+
+////////////////////////////////////////////////////////////////////////////////
+/// Resolve the ADL1 (legacy, context-less) entry points for the Overdrive5
+/// fallback. Only called when ADL2 is unavailable, so these symbols are never
+/// looked up on modern drivers.
+bool profiler::Amd::loadAdl1Symbols()
+{
+    adlMainControlCreate =
+        reinterpret_cast<ADL_MAIN_CONTROL_CREATE>(loadFunction("ADL_Main_Control_Create", false));
+    adlMainControlDestroy =
+        reinterpret_cast<ADL_MAIN_CONTROL_DESTROY>(loadFunction("ADL_Main_Control_Destroy", false));
+    adlAdapterNumberOfAdaptersGet =
+        reinterpret_cast<ADL_ADAPTER_NUMBEROFADAPTERS_GET>(loadFunction("ADL_Adapter_NumberOfAdapters_Get", false));
+    adlAdapterAdapterInfoGet =
+        reinterpret_cast<ADL_ADAPTER_ADAPTERINFO_GET>(loadFunction("ADL_Adapter_AdapterInfo_Get", false));
+    adlOverdrive5CurrentActivityGet =
+        reinterpret_cast<ADL_PM_ACTIVITY_GET>(loadFunction("ADL_Overdrive5_CurrentActivity_Get", false));
+
+    return nullptr != adlMainControlCreate && nullptr != adlOverdrive5CurrentActivityGet;
+}
+
+
 bool profiler::Amd::load()
 {
 #ifdef _WIN32
@@ -58,33 +101,6 @@ bool profiler::Amd::load()
     }
 
     ////////////////////////////////////////////////////////////////////////////
-    // ADL2 (modern) entry points. Optional: absent on very old drivers, in which
-    // case we degrade to the legacy Overdrive5 path below.
-    adl2MainControlCreate =
-        reinterpret_cast<ADL2_MAIN_CONTROL_CREATE>(loadFunction("ADL2_Main_Control_Create", false));
-    adl2MainControlDestroy =
-        reinterpret_cast<ADL2_MAIN_CONTROL_DESTROY>(loadFunction("ADL2_Main_Control_Destroy", false));
-    adl2AdapterNumberOfAdaptersGet =
-        reinterpret_cast<ADL2_ADAPTER_NUMBEROFADAPTERS_GET>(loadFunction("ADL2_Adapter_NumberOfAdapters_Get", false));
-    adl2AdapterAdapterInfoGet =
-        reinterpret_cast<ADL2_ADAPTER_ADAPTERINFO_GET>(loadFunction("ADL2_Adapter_AdapterInfo_Get", false));
-    adl2NewQueryPMLogDataGet =
-        reinterpret_cast<ADL2_PMLOG_QUERY_GET>(loadFunction("ADL2_New_QueryPMLogData_Get", false));
-
-    ////////////////////////////////////////////////////////////////////////////
-    // ADL1 (legacy) entry points, used only for the Overdrive5 fallback.
-    adlMainControlCreate =
-        reinterpret_cast<ADL_MAIN_CONTROL_CREATE>(loadFunction("ADL_Main_Control_Create", false));
-    adlMainControlDestroy =
-        reinterpret_cast<ADL_MAIN_CONTROL_DESTROY>(loadFunction("ADL_Main_Control_Destroy", false));
-    adlAdapterNumberOfAdaptersGet =
-        reinterpret_cast<ADL_ADAPTER_NUMBEROFADAPTERS_GET>(loadFunction("ADL_Adapter_NumberOfAdapters_Get", false));
-    adlAdapterAdapterInfoGet =
-        reinterpret_cast<ADL_ADAPTER_ADAPTERINFO_GET>(loadFunction("ADL_Adapter_AdapterInfo_Get", false));
-    adlOverdrive5CurrentActivityGet =
-        reinterpret_cast<ADL_PM_ACTIVITY_GET>(loadFunction("ADL_Overdrive5_CurrentActivity_Get", false));
-
-    ////////////////////////////////////////////////////////////////////////////
     // ADL's malloc callback. ADL uses it to allocate the buffers it hands back
     // (adapter info, etc), so it MUST honour the requested size — the previous
     // implementation returned a fixed 1-byte block, corrupting the heap the
@@ -95,14 +111,15 @@ bool profiler::Amd::load()
 
     ////////////////////////////////////////////////////////////////////////////
     // Preferred path: ADL2 context + PMLog sensor query. Works on RDNA/Vega and
-    // newer, which is everything the legacy Overdrive5 API cannot read.
-    if (nullptr != adl2MainControlCreate && nullptr != adl2NewQueryPMLogDataGet)
+    // newer, which is everything the legacy Overdrive5 API cannot read. The ADL1
+    // symbols below are only resolved if this path is unavailable.
+    if (true == loadAdl2Symbols())
     {
         if (ADL_OK == adl2MainControlCreate(adlMalloc, 1, &context) && nullptr != context)
         {
+            backend = Backend::PMLOG;
             if (true == buildAdapterMap())
             {
-                backend = Backend::PMLOG;
                 valid = true;
                 logInfo() << "AMD telemetry backend: ADL2 PMLog (" << busToAdapter.size() << " adapter(s))";
                 return true;
@@ -120,16 +137,17 @@ bool profiler::Amd::load()
             adl2MainControlDestroy(context);
         }
         context = nullptr;
+        backend = Backend::NONE;
     }
 
     ////////////////////////////////////////////////////////////////////////////
     // Legacy fallback: ADL1 context + Overdrive5 activity. Pre-GCN1.2 only.
-    if (nullptr != adlMainControlCreate && nullptr != adlOverdrive5CurrentActivityGet)
+    if (true == loadAdl1Symbols())
     {
         if (ADL_OK == adlMainControlCreate(adlMalloc, 1))
         {
-            buildAdapterMap(); // best-effort; Overdrive5 can still use raw indices
             backend = Backend::OVERDRIVE5;
+            buildAdapterMap(); // best-effort; Overdrive5 can still use raw indices
             valid = true;
             logInfo() << "AMD telemetry backend: Overdrive5 (legacy)";
             return true;
@@ -148,7 +166,7 @@ void profiler::Amd::unload()
     {
         adlMainControlDestroy();
     }
-    if (nullptr != context && nullptr != adl2MainControlDestroy)
+    if (Backend::PMLOG == backend && nullptr != context && nullptr != adl2MainControlDestroy)
     {
         adl2MainControlDestroy(context);
         context = nullptr;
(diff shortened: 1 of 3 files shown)


sources/profiler/amd.cpp
```diff
@@ -51,58 +57,303 @@ bool profiler::Amd::load()
...
+        reinterpret_cast<ADL_MAIN_CONTROL_DESTROY>(loadFunction("ADL_Main_Control_Destroy", false));
+    adlAdapterNumberOfAdaptersGet =
+        reinterpret_cast<ADL_ADAPTER_NUMBEROFADAPTERS_GET>(loadFunction("ADL_Adapter_NumberOfAdapters_Get", false));
+    adlAdapterAdapterInfoGet =
+        reinterpret_cast<ADL_ADAPTER_ADAPTERINFO_GET>(loadFunction("ADL_Adapter_AdapterInfo_Get", false));
     adlOverdrive5CurrentActivityGet =
-        reinterpret_cast<ADL_PM_ACTIVITY_GET>(loadFunction("ADL_Overdrive5_CurrentActivity_Get"));
+        reinterpret_cast<ADL_PM_ACTIVITY_GET>(loadFunction("ADL_Overdrive5_CurrentActivity_Get", false));
 
-    IS_NULL(adlMainControlCreate);
```
luminousmining, line 59: We need to check if `adlMainControlCreate == nullptr && adl2MainControlCreate`.
Same logicals with other pointer.

sources/profiler/amd.cpp
```diff
@@ -51,58 +57,303 @@ bool profiler::Amd::load()
...
+        reinterpret_cast<ADL2_MAIN_CONTROL_DESTROY>(loadFunction("ADL2_Main_Control_Destroy", false));
+    adl2AdapterNumberOfAdaptersGet =
+        reinterpret_cast<ADL2_ADAPTER_NUMBEROFADAPTERS_GET>(loadFunction("ADL2_Adapter_NumberOfAdapters_Get", false));
+    adl2AdapterAdapterInfoGet =
+        reinterpret_cast<ADL2_ADAPTER_ADAPTERINFO_GET>(loadFunction("ADL2_Adapter_AdapterInfo_Get", false));
+    adl2NewQueryPMLogDataGet =
+        reinterpret_cast<ADL2_PMLOG_QUERY_GET>(loadFunction("ADL2_New_QueryPMLogData_Get", false));
+
+    ////////////////////////////////////////////////////////////////////////////
+    // ADL1 (legacy) entry points, used only for the Overdrive5 fallback.
```
luminousmining: I think it's unnecessary to load the ADL1 part if ADL2 was successfully loaded!

sources/profiler/amd.cpp
```diff
@@ -51,58 +57,303 @@ bool profiler::Amd::load()
...
+/// adapter index. The ADL adapter index is neither the OpenCL device index nor
+/// the PCI bus, so the only reliable cross-reference is the bus number, which the
+/// device layer already tracks.
+bool profiler::Amd::buildAdapterMap()
 {
+    busToAdapter.clear();
+
+    int numAdapters{ 0 };
+
+    bool const useAdl2{ nullptr != context && nullptr != adl2AdapterNumberOfAdaptersGet
```
luminousmining: Keeping 2 versions loaded and checking the pointers every time is too heavy / complex — we should add an enum to indicate which version is in use: ADL_1, ADL_2.

sources/profiler/amd.cpp
yuzi-co: Reworked so the version is decided by which symbol set actually resolves: `loadAdl2Symbols()` returns true only when the full ADL2 set (create + PMLog query + adapter-info) is present, and `loadAdl1Symbols()` likewise for ADL1. `load()` commits to ADL2 when those resolve and the context creates, otherwise falls back to ADL1 — a single explicit per-version gate instead of scattered per-pointer null checks. (9a5379e)

sources/profiler/amd.cpp
yuzi-co: Done — the ADL1 entry points are resolved inside `loadAdl1Symbols()`, which is only called in the fallback branch after the ADL2 path fails. On a driver where ADL2 works they are never looked up. (9a5379e)

sources/profiler/amd.cpp
yuzi-co: The `Backend` enum (`PMLOG`/`OVERDRIVE5`) is now the single source of truth: `load()` sets it once and `buildAdapterMap()`/`unload()`/`getTelemetry()` dispatch on it instead of re-checking raw pointers each call.

I reused the existing `Backend` enum rather than adding a parallel `ADL_1`/`ADL_2` one — `PMLOG`⇔ADL2 and `OVERDRIVE5`⇔ADL1 are 1:1, so two lock-step enums would reintroduce the complexity you flagged. Happy to add an explicit ADL-version enum instead if you'd prefer.

Verified end-to-end on an RX 9070 XT (RDNA4): the backend selects ADL2 PMLog and reports correct power/clock/utilization. (9a5379e; follow-up 822b7f3 also fixes integrated-GPU adapter mapping discovered during that testing.)

sources/profiler/amd.cpp
```diff
@@ -51,58 +100,294 @@ bool profiler::Amd::load()
...
+bool profiler::Amd::buildAdapterMap()
+{
+    busToAdapter.clear();
+
+    int numAdapters{ 0 };
+
+    // Dispatch on the backend the caller already committed to, not on raw
+    // pointers: PMLOG always means the ADL2 context-based calls, OVERDRIVE5 the
+    // ADL1 context-less ones.
+    if (Backend::PMLOG == backend)
```
luminousmining: For enums, it is preferable to use `switch` statements — this gives us a compiler warning when a new value is added to the enum but some `switch` statements do not handle it.

sources/profiler/amd.cpp
```diff
@@ -51,58 +100,294 @@ bool profiler::Amd::load()
...
+    {
+        return false;
+    }
+
+    if (0 >= numAdapters)
+    {
+        return false;
+    }
+
+    // Value-initialised so ADL sees zeroed structures.
```
luminousmining: You can delete this comment

sources/profiler/amd.cpp
```diff
@@ -51,58 +100,294 @@ bool profiler::Amd::load()
...
+    {
+        return telemetry;
+    }
+
+    int adapterIndex{ 0 };
+    if (false == resolveAdapterIndex(pciBus, adapterIndex))
+    {
+        // Overdrive5 has no adapter map on most paths; fall back to the bus as
+        // index to preserve historical single-GPU behaviour.
+        if (Backend::OVERDRIVE5 != backend)
```
luminousmining: For enums, it is preferable to use switch statements — this gives us a compiler warning when a new value is added to the enum but some switch statements do not handle it.

sources/profiler/amd.cpp
```diff
@@ -51,58 +100,294 @@ bool profiler::Amd::load()
...
+        {
+            if (true == warnedBuses.insert(pciBus).second)
```
luminousmining: I don't want an action and a check to be performed at the same time.
Split the insert and the check into separate steps.

sources/profiler/amd.cpp
```diff
@@ -51,58 +100,294 @@ bool profiler::Amd::load()
...
+bool profiler::Amd::resolveAdapterIndex(uint32_t const pciBus, int& adapterIndex)
+{
+    auto const it{ busToAdapter.find(pciBus) };
+    if (it != busToAdapter.end())
+    {
+        adapterIndex = it->second;
+        return true;
+    }
+
+    if (1 == busToAdapter.size())
```
luminousmining: Maybe `1u` for unsigned

sources/profiler/amd.cpp
```diff
@@ -51,58 +100,294 @@ bool profiler::Amd::load()
...
-                             } };
-    if (ADL_OK != adlMainControlCreate(cbAdlControlCreate, 1))
+    ////////////////////////////////////////////////////////////////////////////
+    // Preferred path: ADL2 context + PMLog sensor query. Works on RDNA/Vega and
+    // newer, which is everything the legacy Overdrive5 API cannot read. The ADL1
+    // symbols below are only resolved if this path is unavailable.
+    if (true == loadAdl2Symbols())
     {
-        return false;
+        if (ADL_OK == adl2MainControlCreate(adlMalloc, 1, &context) && nullptr != context)
```
luminousmining: I prefere separate step.
`if (ADL_OK == adl2MainControlCreate(adlMalloc, 1, &context)`
`if (nullptr != context)`

sources/profiler/amd.cpp
```diff
@@ -51,58 +100,294 @@ bool profiler::Amd::load()
...
+            logWarn() << "ADL PMLog unavailable for adapter " << adapterIndex
+                      << "; telemetry will read zero for this device";
+        }
+        return telemetry;
+    }
+
+    auto const sensor{ [&data](ADL_PMLOG_SENSORS const id) -> int
+                       { return 0 != data.sensors[id].supported ? data.sensors[id].value : 0; } };
+
+    telemetry.coreClock = static_cast<uint32_t>(sensor(ADL_PMLOG_CLK_GFXCLK));
```
luminousmining: use `cast.hpp`

sources/profiler/amd.cpp
```diff
@@ -51,58 +100,294 @@ bool profiler::Amd::load()
...
+        // it negated (-1002 instead of 1002). Match either sign, otherwise the
+        // iGPU is left out of the map and borrows the dGPU's telemetry through the
+        // single-adapter fallback in resolveAdapterIndex().
+        if ((AMD_VENDOR_ID != adapter.iVendorID && -AMD_VENDOR_ID != adapter.iVendorID)
+            || 0 == adapter.iPresent)
+        {
+            continue;
+        }
+        // One physical GPU can expose several ADL indices; keep the first per bus.
+        busToAdapter.emplace(static_cast<uint32_t>(adapter.iBusNumber), adapter.iAdapterIndex);
```
luminousmining: use `cast.hpp`

sources/profiler/amd.cpp
```diff
@@ -51,58 +100,294 @@ bool profiler::Amd::load()
...
+        if (Backend::OVERDRIVE5 != backend)
+        {
+            if (true == warnedBuses.insert(pciBus).second)
+            {
+                logWarn() << "No ADL adapter matches PCI bus " << pciBus
+                          << "; telemetry will read zero for this device";
+            }
+            return telemetry;
+        }
+        adapterIndex = static_cast<int>(pciBus);
```
luminousmining: use `cast.hpp`

sources/profiler/amd.cpp
```diff
@@ -51,58 +100,294 @@ bool profiler::Amd::load()
...
+    int power{ sensor(ADL_PMLOG_BOARD_POWER) };
+    if (0 == power)
+    {
+        power = sensor(ADL_PMLOG_ASIC_POWER);
     }
+    if (0 == power)
+    {
+        power = sensor(ADL_PMLOG_GFX_POWER);
+    }
+    telemetry.power = static_cast<double>(power);
```
luminousmining: use `cast.hpp`

sources/profiler/amd.cpp
```diff
@@ -51,58 +100,294 @@ bool profiler::Amd::load()
...
-    if (ADL_OK != adlOverdrive5CurrentActivityGet(id, &activity))
+    if (ADL_OK != adlOverdrive5CurrentActivityGet(adapterIndex, &activity))
     {
-        logErr() << "ADL cannot get activity";
+        logErr() << "ADL cannot get activity (adapter " << adapterIndex << ")";
+        return telemetry;
     }
 
-    return activity;
+    telemetry.coreClock = static_cast<uint32_t>(activity.iEngineClock) / 100u; // 10 kHz -> MHz
```
luminousmining: use `cast.hpp`

sources/profiler/amd.hpp
```diff
@@ -10,29 +14,88 @@ namespace profiler
     struct Amd
     {
       public:
+        ////////////////////////////////////////////////////////////////////////
+        /// Backend-agnostic telemetry snapshot for a single GPU.
+        struct Telemetry
+        {
+            uint32_t coreClock{ 0u };   //!< Engine/GFX clock (MHz)
```
luminousmining: replace comment `//!<` b `//`

sources/profiler/amd.hpp
```diff
@@ -10,29 +14,88 @@ namespace profiler
...
 
       private:
+        ////////////////////////////////////////////////////////////////////////
+        /// Which AMD telemetry API the loaded driver actually supports. This is
+        /// the single source of truth for the selected ADL version: PMLOG is the
+        /// modern ADL2 path (RDNA/Vega+); OVERDRIVE5 is the legacy ADL1 fallback
+        /// for pre-GCN1.2 cards. Once `load()` picks a backend, every call site
+        /// dispatches on this enum instead of re-checking raw function pointers.
+        /// A future ADLX backend slots in here.
+        enum class Backend
```
luminousmining: Specify the `uint8_t` type.

sources/profiler/amd.cpp
yuzi-co: This is now handled by the backend rework rather than scattered per-pointer checks: `loadAdl2Symbols()` returns true only when the full ADL2 set resolves, `loadAdl1Symbols()` likewise for ADL1, and `load()` commits to a single `Backend`. Every call site dispatches on that enum instead of re-checking raw pointers. (b6625ec)

sources/profiler/amd.cpp
yuzi-co: Done — `buildAdapterMap()` now `switch`es over `Backend` with all three cases and no `default:`, so `-Wswitch` flags a new backend here. (b6625ec)

sources/profiler/amd.cpp
yuzi-co: Done — the bus-fallback in `getTelemetry()` is now an exhaustive `switch (backend)`; the dispatch below it was already a switch. (b6625ec)

sources/profiler/amd.cpp
yuzi-co: Done — split into `bool const firstWarning{ warnedBuses.insert(pciBus).second };` then the check. Applied the same to the `warnedAdapters` insert in `getTelemetryPMLog`. (b6625ec)

sources/profiler/amd.cpp
yuzi-co: Done — extracted `int const createStatus{ adl2MainControlCreate(adlMalloc, 1, &context) };` then `if (ADL_OK == createStatus && nullptr != context)`. (b6625ec)

sources/profiler/amd.cpp
yuzi-co: Done — `if (1u == busToAdapter.size())`. (b6625ec)

sources/profiler/amd.cpp
yuzi-co: Done — removed. (b6625ec)

sources/profiler/amd.cpp
yuzi-co: Done — converted to `cast.hpp` macros (`castU32`/`cast32`/`castSize`/`castDouble`); swept the whole file so no bare `static_cast` remains. (b6625ec)

sources/profiler/amd.cpp
yuzi-co: Done — converted to `cast.hpp` macros (`castU32`/`cast32`/`castSize`/`castDouble`); swept the whole file so no bare `static_cast` remains. (b6625ec)

sources/profiler/amd.cpp
yuzi-co: Done — converted to `cast.hpp` macros (`castU32`/`cast32`/`castSize`/`castDouble`); swept the whole file so no bare `static_cast` remains. (b6625ec)

sources/profiler/amd.cpp
yuzi-co: Done — converted to `cast.hpp` macros (`castU32`/`cast32`/`castSize`/`castDouble`); swept the whole file so no bare `static_cast` remains. (b6625ec)

sources/profiler/amd.cpp
yuzi-co: Done — converted to `cast.hpp` macros (`castU32`/`cast32`/`castSize`/`castDouble`); swept the whole file so no bare `static_cast` remains. (b6625ec)

sources/profiler/amd.hpp
yuzi-co: Done — all trailing comments use plain `//` now. (b6625ec)

sources/profiler/amd.hpp
yuzi-co: Done — `enum class Backend : uint8_t`. (b6625ec)

PR #142, opened 2026-06-07
perf(amd): coalesce progpow LDS cache fill + sub-group barriers for intra-wavefront exchange

yuzi-co:
Two OpenCL kernel optimizations on the AMD progpow/ethash search path. Measured on an RX 9070 XT (RDNA4, gfx1201): ethash +1.9%, kawpow +1.0% (cumulative).

### 1. Coalesce the `initialize_header` LDS cache fill (safe on all hardware)

The old fill stored 4 strided words per lane (AoS), hitting a 4-/8-way LDS bank conflict on every component and an uncoalesced global read. Replaced with a flat grid-stride copy where consecutive lanes write consecutive words:

```c
__global uint const* restrict const dag_words = (__global uint const* restrict const)dag;
for (uint w = thread_id; w < MODULE_CACHE; w += GROUP_SIZE) { header_dag[w] = dag_words[w]; }
```

The LDS layout is **identical** (the dynamic sequence reads `header_dag[hash & (MODULE_CACHE-1)]` at arbitrary offsets, so only the fill *access pattern* changes). Lane `w` → bank `w % 32`, so the store is conflict-free and the global read coalesced. Behaviour-preserving; no hardware assumption.

### 2. `barrier` → `sub_group_barrier` for intra-wavefront lane exchange

The progpow/ethash lane-exchange swaps (`words0`, `reduces`, `share_hash0`, `share_msb_lsb`, …) only exchange data **within** a lane group (`LANES` / `LANE_PARALLEL` consecutive lanes). When that group is contained in a single sub-group (wavefront), a full work-group `barrier` is unnecessary and stalls the other wavefronts on every DAG read; `sub_group_barrier(CLK_LOCAL_MEM_FENCE)` is sufficient.

⚠️ **Needs maintainer validation on non-RDNA4 hardware.** This is correct only while each lane group is contained within one sub-group (verified on RDNA4 where `cl_khr_subgroups` is present and `LANES ≤ wavefront`). If any supported AMD GPU lays out work-groups so a lane group spans sub-groups, this changes results and should be gated. The coalesce change (#1) is independent and safe if you prefer to take only that.

Kernels are loaded at runtime, so this does not affect the build; correctness was checked against the FiroPoW test vectors (see #136) and live share acceptance.

PR #148, opened 2026-06-07
bench(amd): in-tree benchmarks for the ethash, progpow & autolykos search kernels

yuzi-co:
An in-tree A/B benchmark for the AMD `ethash_search` and `progpow_search` kernels, so kernel-level changes (e.g. #142) can be measured against a baseline without a live pool.

### What it adds

- **`sources/benchmark/opencl/ethash/`** and **`.../progpow/`** — `*_baseline.cl` and `*_subgroup.cl`, self-contained copies of each search kernel (baseline vs the `sub_group_barrier` intra-wavefront exchange). The benchmark carries its own copies, so it does **not** depend on the production resolver kernels and stays stable as those evolve.
- **`sources/benchmark/amd/ethash.cpp` / `progpow.cpp`** — drive each kernel variant through the existing benchmark `workflow`, timing the search at a fixed difficulty.
- Wiring: `workflow.{hpp,cpp}` hooks, `config.json` entries, and the OpenCL/algo `CMakeLists.txt` links needed to build the new units.
- `reasearch_and_development/{ethash,kawpow}/amd.md` — the measured RX 9070 XT (RDNA4, gfx1201) A/B numbers (5-run paired), and a one-line note in `BENCHMARK.md`.

### Results (RX 9070 XT)

`sub_group_barrier` vs full work-group `barrier` on the intra-wavefront lane exchange: ethash **+1.3%**, kawpow positive and consistent across paired runs. These corroborate the production change in #142.

Builds under `-DBUILD_EXE_BENCHMARK=ON`; kernels are loaded at runtime so this adds no source-build coupling. Verified to compile + link clean against `main` (clang, `BUILD_AMD=ON`).


yuzi-co:
Added an in-tree **Autolykos v2** throughput benchmark (`runAmdAutolykos`), closing the last AMD coverage gap — autolykos_v2 was the only supported AMD algorithm with no in-tree benchmark.

- Exercises the **production** `autolykos_v2_search` and `autolykos_v2_verify` kernels; defines, append order and launch geometry are byte-for-byte mirrors of `ResolverAmdAutolykosV2`, so the measured kernels are exactly what the miner runs.
- Builds the ~4 GiB DAG once as untimed setup; the displayed grid is the logical nonce grid (`AMD_NONCES_PER_ITER`) so the reported rate is nonces/s.
- No A/B variant here (autolykos has no in-tree optimisation to compare) — the goal is throughput coverage, unlike the ethash/progpow A/B benchmarks.
- Wired into `runAmd()`, `config.json` (disabled by default), and documented in `BENCHMARK.md`.

**Verification (RX 9070 XT / RDNA4, cross-compiled, 20 iters × 3 runs):**
- `autolykos_v2_search`: ~1.0 GH/s steady (0.95–1.5 GH across back-to-back runs; clock-bound), ~2.8–4.4 ms/launch
- `autolykos_v2_verify`: ~67 MH/s steady, ~62 ms/launch
- `ResolverAutolykosv2AmdTest` unit tests pass (2/2).

PR #149, opened 2026-06-08
build: unified Dockerized build matrix — Linux/Windows × AMD/NVIDIA/both (one toolchain)

yuzi-co:
## Unified Dockerized build matrix (Linux host, no local tooling)

Supersedes the earlier per-target cross Dockerfiles. **One `GPU={amd,nvidia,both}` selector × two Dockerfiles** produces every variant — including a **combined AMD+NVIDIA binary** — entirely in Linux containers (Windows is cross-compiled; no Windows host, no engine-mode switch).

| | `GPU=amd` | `GPU=nvidia` | `GPU=both` |
|---|---|---|---|
| `docker/Dockerfile.windows-cross` (PE32+) | OpenCL | CUDA | **AMD+CUDA in one .exe** |
| `docker/Dockerfile.linux` (ELF) | OpenCL | CUDA | AMD+CUDA |

```
scripts/docker-build.ps1 -Os {linux|windows-cross|all} -Gpu {amd|nvidia|both}
# or: docker build -f docker/Dockerfile.windows-cross --build-arg GPU=both --target artifact -o dist/windows-cross-both .
```

### Key design
- **One MSVC ABI** for Windows: AMD moved off mingw (GNU ABI) onto the same clang-cl + xwin toolchain as CUDA, so both backends co-link into a single binary. The mingw AMD path and the six old per-target Dockerfiles are removed.
- **Base chosen by selector:** lean `ubuntu:24.04` for AMD-only; `nvidia/cuda:13.1.2-devel-ubuntu24.04` for nvidia/both. Both Ubuntu 24.04, so the toolchain block is identical on either.
- **OpenSSL** prebuilt MSVC libs (gated on `LM_WIN_CROSS`, not `USE_CLANG_CUDA`, so AMD-only links it too). **OpenCL** on Windows = Khronos ICD loader built from source (vcpkg's `opencl` port compiles under clangcl but fails its `cllayerinfo` tool-copy on cross). vcpkg builds Boost only on windows-cross; the linux image uses vcpkg opencl/openssl natively.

### Validation
All **6 variants build green** (real `docker build`): windows-cross amd/nvidia/both (PE32+ 1.5/12.7/12.9 MB), linux amd/nvidia/both (ELF 11/176/176 MB).

**Runtime-tested on a real AMD host (RX 9070 XT, no NVIDIA):**
- `windows-cross-amd` — enumerates both AMD GPUs, initializes devices, connects stratum, clean exit. ✅
- `windows-cross-both` (combined) — now also runs on the AMD-only host: detects no NVIDIA driver, logs "skipping NVIDIA devices", and mines on AMD. ✅ (Previously crashed in `cudaGetDeviceCount`; fixed by a driver-presence guard before NVIDIA init — `cudart` faults rather than returning an error when no driver is present.)

### Caveats
- **NVIDIA side is build-validated only** (no NVIDIA hardware on hand) — the CUDA kernels compile and link, but haven't been run on an NVIDIA GPU.
- Unpinned `VCPKG_REF`/llvm-mingw/xwin (latest); pin for reproducibility if desired.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


CMakeLists.txt
```diff
@@ -23,6 +24,11 @@ option(BUILD_EXE_BENCHMARK "Build benchmark executable" ON)
 option(TOOL_TRACE_MEMORY "Tool for develloper, trace memory allocation" OFF)
 option(TOOL_MOCKER "Tool for develloper, insert devices fake" OFF)
 option(TOOL_ANALYZER "Tool for develloper, analyzer source code" OFF)
+### DEPENDENCY PROVISIONING
+# When ON, Boost/OpenSSL/OpenCL are resolved through a vcpkg toolchain file
+# (CMAKE_TOOLCHAIN_FILE) instead of the hard-coded C:\Boost / /usr/local paths
+# below. Used by the Docker images and CMake presets.
+option(USE_VCPKG "Resolve Boost/OpenSSL/OpenCL via a vcpkg toolchain instead of hard-coded paths" OFF)
```
luminousmining: Remove listing `Boost/OpenSSL/OpenCL` the files vcpks.json give already the listing.

CMakeLists.txt
```diff
@@ -23,6 +24,11 @@ option(BUILD_EXE_BENCHMARK "Build benchmark executable" ON)
 option(TOOL_TRACE_MEMORY "Tool for develloper, trace memory allocation" OFF)
 option(TOOL_MOCKER "Tool for develloper, insert devices fake" OFF)
 option(TOOL_ANALYZER "Tool for develloper, analyzer source code" OFF)
+### DEPENDENCY PROVISIONING
```
luminousmining: You can delete this comment

CMakeLists.txt
```diff
@@ -112,10 +121,18 @@ endif()
 
 
 if (WIN32)
+    if (MSVC)
```
luminousmining, line 120: indent all the scope please

CMakeLists.txt
```diff
@@ -217,7 +265,73 @@ set(OPEN_SSL_LIBRARIES OpenSSL::SSL OpenSSL::Crypto)
...
+    else()
+        set(_xwin "/opt/xwin")
+    endif()
+    if (DEFINED ENV{CUDA_HOME})
+        set(CLANG_CUDA_PATH "$ENV{CUDA_HOME}")
+    else()
+        set(CLANG_CUDA_PATH "/usr/local/cuda")
+    endif()
+    set(CLANG_CUDA_COMPILER "/opt/llvm-mingw/bin/clang++")
+    set(CLANG_CUDA_GPU_ARCHS sm_75 sm_86 sm_89 sm_120 sm_121)
```
luminousmining: The `NVCC_ARCH_COMPUTE` feature is missing — I think you could either wrap it in a function or handle it upstream so it remains compatible with or without vcpkg. This feature is very useful in the development workflow: it allows compiling for a specific GPU target rather than for all of them, which saves compilation time for developers.

CMakeLists.txt
```diff
@@ -217,7 +265,73 @@ set(OPEN_SSL_LIBRARIES OpenSSL::SSL OpenSSL::Crypto)
...
+            add_custom_command(
+                OUTPUT "${_obj}"
+                # Release device codegen: -O3 -DNDEBUG matches the nvcc release path
+                # (CUDA_NVCC_FLAGS_RELEASE). Without it clang defaults to -O0 and the
+                # mining kernels -- the hot path -- ship unoptimized.
+                COMMAND "${CLANG_CUDA_COMPILER}" --target=x86_64-pc-windows-msvc -x cuda
+                        "${CMAKE_CURRENT_SOURCE_DIR}/${src}" -c -o "${_obj}"
+                        --cuda-path=${CLANG_CUDA_PATH} ${_archflags}
+                        -O3 -DNDEBUG
+                        -Wno-unknown-cuda-version -std=c++20 -fms-compatibility -fms-extensions
```
luminousmining: Can you define 1 flag by line please it's more readable

CMakeLists.txt
```diff
@@ -273,22 +387,84 @@ if (BUILD_NVIDIA)
...
 ### FIND PACKAGE | LIBRARIES                                                ####
 ################################################################################
-find_package(Boost 1.91.0 EXACT COMPONENTS atomic chrono filesystem system json thread serialization program_options)
+if (USE_VCPKG)
+    # vcpkg ships Boost as a CONFIG package; its exact version follows the vcpkg
+    # baseline (see vcpkg.json) rather than being pinned to 1.91.0 here.
+    find_package(Boost CONFIG REQUIRED COMPONENTS atomic chrono filesystem system json thread serialization program_options)
+    if (NOT Boost_LIBRARIES)
+        set(Boost_LIBRARIES
+            Boost::atomic Boost::chrono Boost::filesystem Boost::system
```
luminousmining: If you can not list all lib on 1 line, can you write 1 lib by line please

CMakeLists.txt
```diff
@@ -273,22 +387,84 @@ if (BUILD_NVIDIA)
...
+endif()
 if (UNIX)
     find_package(GnuTLS)
 endif()
-find_library(OpenCL_LIBRARIES name OpenCL PATHS ${OpenCL_LIBRARY_DIR} PATH_SUFFIXES lib REQUIRED)
-if (BUILD_NVIDIA)
+# Only needed for AMD builds, and only if find_package(OpenCL) above didn't
+# already populate OpenCL_LIBRARIES (it does under vcpkg). The previous
+# unconditional form also forced an OpenCL dependency onto NVIDIA/CPU-only builds.
+if (BUILD_AMD AND NOT OpenCL_LIBRARIES)
```
luminousmining, line 480: There is probably a confusion with `USE_VCPKG`? If so, you can remove the comment above.
If not, in which case would `OpenCL_LIBRARIES` not be found by vcpkg?

CMakePresets.json
```diff
@@ -0,0 +1,55 @@
...
+      "inherits": "vcpkg-base",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build/windows-cross",
+      "cacheVariables": {
+        "BUILD_AMD": "ON",
+        "BUILD_NVIDIA": "ON",
+        "BUILD_CPU": "OFF",
+        "USE_CLANG_CUDA": "ON",
+        "VCPKG_MANIFEST_NO_DEFAULT_FEATURES": "ON",
+        "VCPKG_MANIFEST_FEATURES": "",
```
luminousmining, line 44: Why `opencl;openssl` are missing ?

vcpkg.json
```diff
@@ -0,0 +1,44 @@
...
+  ],
+  "features": {
+    "opencl": {
+      "description": "OpenCL ICD loader (AMD GPU path).",
+      "dependencies": [
+        "opencl"
+      ]
+    },
+    "openssl": {
+      "description": "OpenSSL via vcpkg (stratum TLS). Omitted on windows-cross (prebuilt MSVC libs); used on linux.",
```
luminousmining: OpenSSL should be enable on Windows also ?

CMakeLists.txt
yuzi-co: Done in b097539 — removed the comment block and the `### DEPENDENCY PROVISIONING` banner.

CMakeLists.txt
yuzi-co: Done in b097539 — trimmed the option string to `"Resolve dependencies via a vcpkg toolchain instead of hard-coded paths"` so it no longer enumerates the libs (vcpkg.json is the source of truth).

CMakeLists.txt
yuzi-co: Done in b097539 — indented the `if (MSVC)` and non-MSVC branch bodies so they nest visibly under `if (WIN32)`.

CMakeLists.txt
yuzi-co: Good catch — restored on the clang-CUDA cross path in b097539. It now mirrors the nvcc branch: if `NVCC_ARCH_COMPUTE` is set the cross build compiles only `sm_${NVCC_ARCH_COMPUTE}`, otherwise the full default arch set. So the single-target dev workflow works on both the nvcc and clang-CUDA paths.

CMakeLists.txt
yuzi-co: Done in b097539 — one flag per line.

CMakeLists.txt
yuzi-co: Done in b097539 — one lib per line.

CMakeLists.txt
yuzi-co: No `USE_VCPKG` confusion — the comment was just unclear, reworded in b097539. `find_package(OpenCL)` populates `OpenCL_LIBRARIES` under vcpkg; on the non-vcpkg hard-coded-path build it may set only the include dir, so the manual `find_library` is the fallback for that case. It is scoped to AMD because the previous unconditional form forced an OpenCL dependency onto NVIDIA/CPU-only builds.

CMakePresets.json
yuzi-co: On windows-cross both are provisioned outside vcpkg, so the manifest features are intentionally empty: OpenSSL cannot be cross-built to the MSVC ABI by vcpkg, so it is staged as prebuilt MSVC libs in the image; and vcpkg's `opencl` port fails its `cllayerinfo` tool-copy under cross-compilation, so the Dockerfile builds the Khronos ICD loader directly and points CMake at it via `-DOpenCL_LIBRARY`/`-DOpenCL_INCLUDE_DIR`. vcpkg therefore supplies only Boost here. The `linux` preset uses vcpkg for both, hence `opencl;openssl`. Added a `description` to the preset spelling this out (b097539).

vcpkg.json
yuzi-co: It is — just not through vcpkg. The windows-cross image links prebuilt MSVC OpenSSL (FireDaemon) staged in the Dockerfile and ships the DLLs in the artifact, because vcpkg cannot cross-build OpenSSL to the MSVC ABI from Linux. The `openssl` vcpkg feature is for the native/linux path. Clarified the wording in the manifest (b097539).

PR #151, opened 2026-06-08
fix(stratum): derive ProgPoW DAG epoch from seed hash, block number for Quai

yuzi-co:
## Problem

progpow-quai fails to mine: the DAG build is rejected with `CL_INVALID_BUFFER_SIZE` → `UPDATE_MEMORY_FAIL`. The root cause is an **inflated DAG epoch**, not a legitimately oversized DAG (an RX 9070 XT's `CL_DEVICE_MAX_MEM_ALLOC_SIZE` is ~13.5 GiB, well above a correct ~4.7 GiB Quai DAG).

The EthereumV1 path derived the DAG epoch from `blockNumber / EPOCH_LENGTH`. That is wrong for **FiroPoW**, whose epoch length changed across the chain's history — `blockNumber/EPOCH_LENGTH` then yields the wrong epoch and a wrong DAG (`Invalid Mixhash` rejects on live Firo). The network **seed hash** encodes the authoritative epoch, so this change prefers it (falling back to `blockNumber/EPOCH_LENGTH` only when the seed isn't recognized).

**progpow-quai is the exception.** Its pool seed hash is keccak-iterated on a much shorter cadence than its DAG-size epoch, so `findEpoch` returns a wildly inflated number — **586** vs the real **11** at block `4401536` — and the DAG balloons to ~40 GiB. The official [quai-gpu-miner](https://github.com/dominant-strategies/quai-gpu-miner) derives the epoch from `blockNumber / EPOCH_LENGTH` (`EPOCH_LENGTH = 388800` in `libprogpow/ProgPow.h`) and ignores the seed hash, so Quai opts into that via `deriveEpochFromBlockNumber`.

Well-behaved coins (kawpow / meowpow / evrprogpow) are unaffected: their seed hash and `blockNumber/EPOCH_LENGTH` agree, so seed-first returns the same epoch.

## Change

- Add `StratumProgPOW::deriveEpochFromBlockNumber` (default `false`).
- V1 path now prefers the seed hash, except when `deriveEpochFromBlockNumber` is set (Quai), which uses `blockNumber / maxEpochLength`.
- `StratumProgpowQuai` sets the flag.
- Log the derived epoch + DAG size in `ResolverAmdProgPOW::updateContext`, so a bad epoch is visible before allocation.

## Verification

- progpow-quai kernel constants checked against go-quai / quai-gpu-miner and already correct: `datasetInitBytes 1<<32`, `datasetGrowthBytes 1<<26`, `datasetParents 512`, `PROGPOW_PERIOD 10`, `PROGPOW_CNT_CACHE 11`, `PROGPOW_CNT_MATH 18`.
- **Live on an RX 9070 XT (gfx1201):** `DAG epoch 11 | light cache 39844928 B | DAG 5033164672 B (4799 MiB)`, no CL error, full-speed mining. Light-cache build dropped from ~33 s (the inflated epoch 586) to ~0.9 s (epoch 11).
- **Hermetic tests** (`sources/stratum/tests/progpow_quai.cpp`, no GPU): Quai derives epoch 11; base ProgPOW still recovers 586 from the seed hash (FiroPoW non-regression). Built and passing via the Docker cross-compile (mingw + clang) unit-test path.

## Note

`quai.herominers.com:1185` turned out to be a **KawPow** pool (seed-hash epoch length ~7500), confirmed by `--algo=kawpow` producing accepted shares there. So progpow-quai's DAG build is verified correct, but an end-to-end progpow-quai *share* test still needs a confirmed progpow-quai endpoint.


sources/stratum/tests/progpow_seed.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+TEST(StratumProgpowQuaiEpoch, baseProgpowStillPrefersSeedHash)
+{
+    stratum::StratumProgPOW stratum{};
+    stratum.stratumType = stratum::STRATUM_TYPE::ETHEREUM_V1;
+
+    stratum.onMiningNotify(makeQuaiNotify());
+
+    EXPECT_EQ(586, stratum.jobInfo.epoch);
+}
```
luminousmining, line 98: Add new line please

sources/stratum/progpow.cpp
```diff
@@ -317,14 +317,31 @@ void stratum::StratumProgPOW::onMiningNotify(boost::json::object const& root)
             }
 
             ////////////////////////////////////////////////////////////////////
-            int32_t epoch{ 0 };
-            if (jobInfo.blockNumber > 0ull)
+            // The DAG epoch is whatever the network seedHash encodes — authoritative
```
luminousmining: Duplicate comments

```
        // Quai decouples the network seed hash from the DAG-size epoch: its pool
        // seed hash is keccak-iterated far more than blockNumber/EPOCH_LENGTH, so
        // matching the seed (findEpoch) yields a hugely inflated epoch and an
        // un-allocatable multi-tens-of-GiB DAG. The official quai-gpu-miner derives
        // the epoch from blockNumber/EPOCH_LENGTH and ignores the seed hash. Set
        // this for such chains; leave false to keep the seed-first derivation that
        // FiroPoW relies on (its EPOCH_LENGTH changed across the chain's history).
        bool deriveEpochFromBlockNumber{ false };
```

Keep only the comment from `deriveEpochFromBlockNumber`

sources/stratum/progpow.hpp
```diff
@@ -26,6 +26,15 @@ namespace stratum
         uint32_t maxEthashEpoch{ algo::ethash::EPOCH_LENGTH };
         uint32_t maxEpochLength{ algo::progpow::EPOCH_LENGTH };
 
+        // Quai decouples the network seed hash from the DAG-size epoch: its pool
```
luminousmining: You mention FiroPoW specifically, but shouldn't this apply to ProgPoW and its derivatives in general?

sources/resolver/amd/progpow.cpp
```diff
@@ -64,6 +64,13 @@ bool resolver::ResolverAmdProgPOW::updateContext(stratum::StratumJobInfo const&
         return false;
     }
 
+    // Surface the derived epoch and DAG footprint: a wrong epoch (e.g. a seed-hash
```
luminousmining: You can delete this part of code.

sources/stratum/firopow.hpp
```diff
@@ -12,5 +12,8 @@ namespace stratum
       public:
         StratumFiroPOW();
         ~StratumFiroPOW() = default;
+
+      protected:
+        int32_t deriveEpoch(stratum::StratumJobInfo const& jobInfo) const override;
```
luminousmining: `override` -> `final`

sources/stratum/progpow.hpp
```diff
@@ -26,6 +26,12 @@ namespace stratum
         uint32_t maxEthashEpoch{ algo::ethash::EPOCH_LENGTH };
         uint32_t maxEpochLength{ algo::progpow::EPOCH_LENGTH };
 
+        // Derive the DAG epoch for the current job. The base maps the block number
```
luminousmining: You can delete this comment, the implementation have already good comments

sources/stratum/tests/progpow_epoch.cpp
```diff
@@ -0,0 +1,73 @@
...
+            true,
+            4401536,
+            "1b01fa91"
+        };
+        return root;
+    }
+}
+
+
+TEST(StratumProgpowQuaiEpoch, derivesEpochFromBlockNumberNotSeedHash)
```
luminousmining: File name is `progpow_seed.cpp` so the Class of the test should be `StratumProgpowSeed` and the name `quaiEpoch`.

sources/stratum/tests/progpow_epoch.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+    stratum.onMiningNotify(makeQuaiNotify());
+
+    EXPECT_EQ(4401536ull, stratum.jobInfo.blockNumber);
+    EXPECT_EQ(static_cast<int32_t>(4401536ull / algo::progpow_quai::EPOCH_LENGTH), stratum.jobInfo.epoch);
+    EXPECT_EQ(11, stratum.jobInfo.epoch);
+}
+
+
+TEST(StratumFiroPowEpoch, prefersSeedHashOverBlockNumber)
```
luminousmining: File name is `progpow_seed.cpp` so the Class of the test should be `StratumProgpowSeed` and the name `firopowEpoch`.

sources/stratum/progpow.hpp
```diff
@@ -23,9 +23,15 @@ namespace stratum
 
       protected:
         uint32_t maxPeriod{ algo::progpow::v_0_9_2::MAX_PERIOD };
-        uint32_t maxEthashEpoch{ algo::ethash::EPOCH_LENGTH };
+        // Upper bound for the seedHash -> epoch search (a max epoch *number*, not a
+        // per-epoch block count). Generous on purpose so the search covers every
+        // realistic progpow/FiroPoW epoch; see deriveEpoch.
+        uint32_t maxEpochSearch{ algo::ethash::EPOCH_LENGTH };
```
luminousmining: Why was `maxEthashEpoch` renamed to `maxEpochSearch`? Both variables have the same type and the same default value (`algo::ethash::EPOCH_LENGTH`), so the distinction is unclear. If the semantics are identical, the rename adds confusion rather than clarity — what is the reasoning behind it?

sources/stratum/progpow.hpp
```diff
@@ -23,9 +23,15 @@ namespace stratum
 
       protected:
         uint32_t maxPeriod{ algo::progpow::v_0_9_2::MAX_PERIOD };
-        uint32_t maxEthashEpoch{ algo::ethash::EPOCH_LENGTH };
+        // Upper bound for the seedHash -> epoch search (a max epoch *number*, not a
```
luminousmining: You can remove the comment — the code is self-explanatory.

sources/stratum/progpow.hpp
```diff
@@ -23,9 +23,15 @@ namespace stratum
...
+        // per-epoch block count). Generous on purpose so the search covers every
+        // realistic progpow/FiroPoW epoch; see deriveEpoch.
+        uint32_t maxEpochSearch{ algo::ethash::EPOCH_LENGTH };
+        // Blocks-per-epoch divisor for the blockNumber/length mapping.
```
luminousmining: You can remove the comment — the code is self-explanatory.

sources/stratum/progpow.hpp
yuzi-co: Good call — reverted to `maxEthashEpoch` and dropped the member comments (2ebab9d). The rename came from folding in #140; the intent was only to signal it's the seed→epoch search *ceiling* (a max epoch number) rather than ethash-specific, distinct from `maxEpochLength` (the blocks-per-epoch divisor). But you're right that it doesn't earn the churn, so I've kept the original name.

PR #154, opened 2026-06-08
feat: add kHeavyHash (Kaspa) algorithm

yuzi-co:
Adds **kHeavyHash** (Kaspa proof-of-work) as a new non-DAG algorithm, built additively across all five layers. Correctness was the hard requirement throughout: every layer is gated bit-identical against the rusty-kaspa reference via known-answer tests, and the end result was confirmed with **real pool-accepted shares** on an RX 9070 XT.

## What's included

- **CPU reference** (`sources/algo/kheavyhash/`) — matrix generation (xoshiro256++ + full-rank gate), keccak-f[1600], PowHash/HeavyHash (cSHAKE256), heavy matmul step, `meetsTarget`. Standalone (no Boost/CUDA/OpenCL) so it doubles as the correctness oracle.
- **OpenCL kernel** (`opencl/kheavyhash.cl`) — bit-identical to the CPU reference; host generates the 64×64 nibble matrix and uploads it (kernel does no rank-check). Ships an optimized `search` variant — see *AMD OpenCL optimization* below.
- **AMD resolver** (`resolver/amd/`) — no DAG; per-job matrix/header/target upload, 1-D `search` launch.
- **CUDA kernel + NVIDIA resolver** (`algo/kheavyhash/cuda/`, `resolver/nvidia/`) — ported from the OpenCL path, modeled on the blake3 (non-DAG) resolver.
- **Stratum client** (`stratum/kheavyhash.cpp`) — Kaspa gostratum dialect; handles both the NORMAL (`[jobId,[4 LE u64],ts]`) and BIG (`[jobId,"<80 hex>"]`) job encodings, `set_extranonce`, and difficulty→target conversion.
- **Benchmark** (`sources/benchmark/amd/kheavyhash.cpp`) — runs the reference plus every optimized variant back-to-back through the `bin/benchmark` framework; results + RGA static analysis in `BENCH.md`.

## AMD OpenCL optimization

The shipped kernel is **`search_lm4`** — **393 → 575 MH/s (+46%)** on an RX 9070 XT (RDNA4 / gfx1201). Five variants live in the same `kheavyhash.cl`, selected by name, each **bit-identical to the reference** (OpenCL KAT, run on the GPU):

| kernel | MH/s | vs ref | optimization |
|---|---|---|---|
| `search` | 393 | 1.00× | reference |
| `search_lm1` | 412 | 1.04× | matrix staged in LDS |
| `search_lm2` | 482 | 1.22× | + `v_dot4_u32_u8` nibble matmul |
| `search_lm3` | 481 | 1.22× | + unrolled keccak (no gain) |
| **`search_lm4`** | **575** | **1.46×** | + powHash keccak midstate ← **shipped** |
| `search_lm5` | 575 | 1.46× | + heavy-keccak round-0 hoist (ties lm4) |

Two levers carried it: the **`v_dot4` matmul** (gfx1201 exposes no `cl_khr_integer_dot_product`, but `__builtin_amdgcn_udot4` lowers to `v_dot4_u32_u8`; guarded by `__AMDGCN__` with a bit-identical scalar fallback for POCL/CI) and the **powHash keccak midstate** (only `state[9]`=nonce varies per nonce, so round-1 is hoisted once per workgroup to LDS). The kernel is compute/keccak-bound, not occupancy-bound (lm1 hit max occupancy for only +5%). `search_lm5` confirms the heavy second keccak is at its floor — the matmul output fills 4 of 5 columns, so nothing hoists past round-0 theta — hence the ~1.46× ceiling. `lm3`/`lm5` are retained in-tree as recorded negative results.

## Verification

- **CPU KATs** 8/8, **OpenCL KATs** — per-stage plus all 6 `search` variants gated bit-identical to the reference (POCL in CI, real RX 9070 XT on the rig), **stratum-math KATs** 4/4, **CUDA host KATs** 5/5 — all against rusty-kaspa vectors.
- **Live**: 3 pool-accepted shares on a real RX 9070 XT (RDNA4), 0 rejects.
- **AMD throughput**: ~575 MH/s shipped (`search_lm4`), reference `search` ~393 MH/s. Full table + RGA (VGPR / occupancy / `v_dot4` / spills) in `BENCH.md`.
- The CUDA path is host-KAT-verified and clang-CUDA compile-verified; it has **not** been run on NVIDIA hardware (none available). The optimization pass is AMD/OpenCL only.

## Notes for review

- Fully **additive and isolated** — new files plus thin integration touchpoints in `device.cpp`, `stratums.cpp`, `job_info`, and the benchmark wiring. No existing algorithm is affected.
- CMake for `algo/kheavyhash/` is guarded so the CPU/OpenCL reference can also build standalone (dependency-free) for KAT verification.


sources/algo/kheavyhash/matrix.cpp
```diff
@@ -0,0 +1,95 @@
...
+                        }
+                    }
+                }
+            }
+        }
+        return rank;
+    }
+
+
+    namespace
```
luminousmining: why insert namespace `here`, you can remove it

sources/algo/kheavyhash/matrix.cpp
```diff
@@ -0,0 +1,95 @@
...
+    {
+        Matrix randMatrixNoRankCheck(Xoshiro256pp& generator)
+        {
+            Matrix mat{};
+            for (size_t i{ 0 }; i < 64; ++i)
```
luminousmining: missing prefix `u`

sources/algo/kheavyhash/matrix.cpp
```diff
@@ -0,0 +1,95 @@
...
+            return mat;
+        }
+    }
+
+
+    Matrix generateMatrix(Hash256 const& seed)
+    {
+        // Regenerate from the SAME continuing stream until full rank, per matrix.rs::generate.
+        Xoshiro256pp generator{ seed };
+        while (true)
```
luminousmining: Is there no maximum loop limit?
How do we detect an infinite loop?

documentation/BENCHMARK.md
```diff
@@ -369,3 +372,23 @@ Measured on 2026-03-16, 10 iterations per kernel, config: `sources/benchmark/con
...
 | lm10 | ~16.25 MH | 64 blocks only — warp parallelism, needs tuning |
 | lm11 | ~27.35 MH | |
+
+---
+
+## Reference Results — AMD Radeon RX 9070 XT (RDNA4 / gfx1201)
+
+200 iterations per kernel, steady-state median (first 40 ramp iterations dropped),
+idle GPU. All six kernels are gated bit-identical by the OpenCL KAT. See the
+[kHeavyHash AMD study](reasearch_and_development/kheavyhash/amd.md) for the full
```
luminousmining, line 396: This reference doest not match with title `Reference Results — AMD Radeon RX 9070 XT (RDNA4 / gfx1201)` you should pace this information on `kHeavyHash` scope.

sources/algo/kheavyhash/cuda/kheavyhash_device.cuh
```diff
@@ -0,0 +1,210 @@
...
+// This is a faithful port of sources/algo/kheavyhash/opencl/kheavyhash.cl, which
+// is itself KAT-verified bit-identical to the CPU reference. The constants are
+// copied verbatim from that kernel (and the CPU reference). It is deliberately
+// written in plain C with __CUDACC__ shims so the SAME source can be compiled on
+// the host and checked against the known-answer vectors (tests/cuda_device_test.cpp)
+// even on a machine without an NVIDIA GPU. Only <cstdint> is required on the host.
+
+#include <cstdint>
+
+#if defined(__CUDACC__)
```
luminousmining: I will take the simplification of:

```
#if defined(__CUDACC__)
#define KH_FN __device__ __forceinline__
#define KH_CONST __device__ __constant__
#else
#define KH_FN inline
#define KH_CONST static const
#endif
```

But I prefer to have a dedicated CUDA kernel and pure `cpu` functions, because in the future we might optimize the CPU code using direct `asm`, which would no longer be easily compatible with `cuda`. So for this `.cuh` file, we stay strictly in `cuda` format.

Replace directly with `__device__ __forceinline__` and `__device__ __constant__`.

sources/algo/kheavyhash/cuda/kheavyhash_device.cuh
```diff
@@ -0,0 +1,210 @@
...
+        {
+            for (int b = 0; b < 8; ++b)
+            {
+                out[w * 8 + b] = (uint8_t)((state[w] >> (8 * b)) & 0xFF);
+            }
+        }
+    }
+
+
+    KH_FN void keccakF1600(uint64_t* a)
```
luminousmining: uint64_t* const a

sources/algo/kheavyhash/cuda/kheavyhash_device.cuh
```diff
@@ -0,0 +1,210 @@
...
+        for (int i = 0; i < 32; ++i)
+        {
+            product[i] ^= hash1[i];
+        }
+        kHeavyHash(product, out);
+    }
+
+
+    // pow <= target as little-endian 256-bit integers (scan from MSB).
+    KH_FN bool meetsTarget(uint8_t const* powLe, uint8_t const* targetLe)
```
luminousmining: missing `__restrict__` ?

sources/algo/kheavyhash/cuda/kheavyhash_device.cuh
```diff
@@ -0,0 +1,210 @@
...
+            state[w] ^= loadLe64(input + w * 8);
+        }
+        keccakF1600(state);
+        storeLe256(state, out);
+    }
+
+
+    // Heavy step: matrix * nibble-vector, two rows collapse to one byte (>>10),
+    // XOR with hash1, then KHeavyHash. matrix is row-major uint16_t[64*64], 0..15.
+    KH_FN void heavyHash(uint16_t const* matrix, uint8_t const* hash1, uint8_t* out)
```
luminousmining: missing `__restrict__` ?

sources/algo/kheavyhash/cuda/kheavyhash_device.cuh
```diff
@@ -0,0 +1,210 @@
...
+        for (int i = 0; i < 32; ++i)
+        {
+            uint16_t sum1 = 0;
+            uint16_t sum2 = 0;
+            for (int j = 0; j < 64; ++j)
+            {
+                sum1 = (uint16_t)(sum1 + matrix[(2 * i) * 64 + j] * vec[j]);
+                sum2 = (uint16_t)(sum2 + matrix[(2 * i + 1) * 64 + j] * vec[j]);
+            }
+            product[i] = (uint8_t)(((sum1 >> 10) << 4) | (sum2 >> 10));
```
luminousmining: Explicit:
```cpp
uint8_t const high_nibble = (sum1 >> 10) & 0xF;
uint8_t const low_nibble  = (sum2 >> 10) & 0xF;
product[i] = (high_nibble << 4) | low_nibble;
```

sources/algo/kheavyhash/cuda/kheavyhash_device.cuh
```diff
@@ -0,0 +1,210 @@
...
+            product[i] ^= hash1[i];
+        }
+        kHeavyHash(product, out);
+    }
+
+
+    // pow <= target as little-endian 256-bit integers (scan from MSB).
+    KH_FN bool meetsTarget(uint8_t const* powLe, uint8_t const* targetLe)
+    {
+        for (int i = 31; i >= 0; --i)
```
luminousmining: We should create common function in file `compare.cuh`
```cpp
template <typename T, uint32_t N>
__forceinline__ __device__
bool isLowerLittleEndian(
    T const* __restrict__ r,
    T const* __restrict__ l)
{
}
```

sources/algo/kheavyhash/cuda/memory.cuh
```diff
@@ -0,0 +1,57 @@
...
+}
+
+
+__host__
+bool kheavyhashUpdateConstants(
+    resolver::nvidia::kheavyhash::KernelParameters& params)
+{
+    CUDA_ER(cudaMemcpy(params.matrix,
+                       params.hostMatrix,
+                       64u * 64u * sizeof(uint16_t),
```
luminousmining: 64u -> magic number, we need `constexpr`

sources/algo/kheavyhash/opencl/tests/opencl_kat_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+#include <gtest/gtest.h>
+
+#include <algo/kheavyhash/matrix.hpp>
+#include <algo/kheavyhash/result.hpp>
+#include <common/error/opencl_error.hpp>
+#include <common/kernel_generator/opencl.hpp>
+#include "kheavyhash_test_vectors.hpp"
+
+
+namespace
```
luminousmining: no namespace empty

sources/algo/kheavyhash/opencl/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+# Deploy the kHeavyHash OpenCL kernel next to the binary so the AMD resolver
```
luminousmining: you can delete comments

sources/algo/kheavyhash/opencl/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+    COMMENT "Copy ${FILE}"
+  )
+endforeach()
+
+add_custom_target(copy_kheavyhash_opencl_files ALL
+  DEPENDS
+    ${OUT_KHEAVYHASH}/kheavyhash.cl
+)
+
+# Fold the OpenCL kHeavyHash KAT into the main unit_test on the AMD path so the
```
luminousmining: you can delete comment

sources/algo/kheavyhash/tests/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+# Fold the kHeavyHash CPU known-answer tests into the main unit_test target.
```
luminousmining: delete comments on this file

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
```diff
@@ -0,0 +1,256 @@
+// kHeavyHash (Kaspa) OpenCL kernels.
```
luminousmining: inline indent - all file .cl please

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
```diff
@@ -0,0 +1,256 @@
...
+    0x8000000080008081UL, 0x8000000000008080UL, 0x0000000080000001UL, 0x8000000080008008UL };
+
+__constant int ROTATIONS[24] = { 1, 3, 6, 10, 15, 21, 28, 36, 45, 55, 2, 14,
+                                 27, 41, 56, 8, 25, 43, 62, 18, 39, 61, 20, 44 };
+
+__constant int PI_LANE[24] = { 10, 7, 11, 17, 18, 3, 5, 16, 8, 21, 24, 4,
+                               15, 23, 19, 13, 12, 2, 20, 14, 22, 9, 6, 1 };
+
+
+inline ulong rotl64(ulong const x, int const k)
```
luminousmining: Already exist in `rotate_byte.cl`

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
```diff
@@ -0,0 +1,256 @@
...
+{
+    return (x << k) | (x >> (64 - k));
+}
+
+
+inline ulong loadLe64(uchar const* p)
```
luminousmining: Should be common fonction in common/opencl

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
```diff
@@ -0,0 +1,256 @@
...
+    ulong v = 0;
+    for (int b = 0; b < 8; ++b)
+    {
+        v |= ((ulong)p[b]) << (8 * b);
+    }
+    return v;
+}
+
+
+inline void storeLe256(ulong const* state, uchar* out)
```
luminousmining: Should be common fonction in common/opencl

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
```diff
@@ -0,0 +1,256 @@
...
+    {
+        for (int b = 0; b < 8; ++b)
+        {
+            out[w * 8 + b] = (uchar)((state[w] >> (8 * b)) & 0xFF);
+        }
+    }
+}
+
+
+void keccakF1600(ulong* a)
```
luminousmining: already exist `crypto/opencl/keccak_f1600.cl`

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
```diff
@@ -0,0 +1,256 @@
...
+            return powLe[i] < targetLe[i];
+        }
+    }
+    return true;
+}
+
+
+
+
+// Result buffer shared with the host (mirrors algo::ethash/blake3 Result).
```
luminousmining: Write only by KernelGenerator

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
```diff
@@ -0,0 +1,256 @@
...
+// MAX_RESULT is overridable by the host kernel generator (addDefine).
+#ifndef MAX_RESULT
+#define MAX_RESULT 4
+#endif
+
+typedef struct __attribute__((aligned(8)))
```
luminousmining: benchmark/common/result.cl

For benchmark we doest not need 4 nonces only 1 nonce found is good !

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
```diff
@@ -0,0 +1,256 @@
+// kHeavyHash (Kaspa) OpenCL kernels.
+//
```
luminousmining, line 2: Many remarks in this file, apply for other kernel OpenCL.

sources/algo/kheavyhash/cuda/memory.cuh
```diff
@@ -0,0 +1,57 @@
...
+    resolver::nvidia::kheavyhash::KernelParameters& params)
+{
+    ////////////////////////////////////////////////////////////////////////////
+    if (false == kheavyhashFreeMemory(params))
+    {
+        return false;
+    }
+
+    ////////////////////////////////////////////////////////////////////////////
+    CU_ALLOC(&params.matrix, 64u * 64u * sizeof(uint16_t));
```
luminousmining: Magic number

sources/algo/kheavyhash/cuda/memory.cuh
```diff
@@ -0,0 +1,57 @@
...
+    CU_ALLOC(&params.header, algo::LEN_HASH_256);
+    CU_ALLOC(&params.target, algo::LEN_HASH_256);
+    CU_ALLOC_HOST(&params.resultCache, sizeof(algo::kheavyhash::Result) * 2u);
```
luminousmining: magic number `2u`

sources/algo/kheavyhash/opencl/tests/opencl_kat_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+    // compile-time in-tree source path (POCL/dev), then the deployed path next to the
+    // binary. First that opens wins, so the same unit_test runs in the dev harness and on
+    // a real GPU. Returns "" if none open.
+    std::string resolveKernelPath()
+    {
+        char const* const candidates[3]{ std::getenv("KH_CL_PATH"), KH_CL_PATH,
+                                         "kernel/kheavyhash/kheavyhash.cl" };
+        for (char const* const cand : candidates)
+        {
+            if (nullptr != cand && std::ifstream{ cand }.good())
```
luminousmining: `std::ifstream{ cand }.good()` need 2 steps

sources/algo/kheavyhash/opencl/tests/opencl_kat_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+        for (int i{ 0 }; i < 32; ++i)
+        {
+            h[i] = a[i];
+        }
+        return h;
+    }
+}
+
+
+TEST_F(OpenClKat, PowHashMatchesReference)
```
luminousmining: `PowHashMatchesReference` is a function name so lowerCase

sources/algo/kheavyhash/opencl/tests/opencl_kat_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+{
+    try
+    {
+        std::array<uint8_t, 32> pre{};
```
luminousmining: magic number, I think is `HASH_256_LEN` ?

sources/algo/kheavyhash/opencl/tests/opencl_kat_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+// whose nonce reproduces FP_FINAL must report a hit at exactly FP_NONCE, and must NOT
+// report one against a target one unit below FP_FINAL.
+class SearchKernel : public OpenClKat, public ::testing::WithParamInterface<char const*>
+{
+  protected:
+    // Mirrors the kernel's Result struct (and algo::ethash::Result layout).
+    struct alignas(8) Result
+    {
+        uint8_t  found{ 0 };
+        uint32_t count{ 0 };
```
luminousmining: 0u

sources/algo/kheavyhash/opencl/tests/opencl_kat_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+        uint64_t nonces[4]{ 0, 0, 0, 0 };
```
luminousmining: 0ull

sources/algo/kheavyhash/tests/cuda_device_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+            {
+                f.push_back(m[r][c]);
+            }
+        }
+        return f;
+    }
+}
+
+
+TEST(CudaDeviceKat, PowHashMatchesReference)
```
luminousmining: `PowHashMatchesReference` function name lowerCase

sources/algo/kheavyhash/tests/cuda_device_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+    uint8_t out[32];
+    kheavyhash_cuda::kHeavyHash(kheavyhash::kat::HEAVY_INPUT.data(), out);
+    EXPECT_EQ(toArr(out), kheavyhash::kat::KHEAVY_EXPECTED);
+}
+
+
+TEST(CudaDeviceKat, HeavyHashMatchesReference)
+{
+    std::vector<uint16_t> flat;
+    flat.reserve(64 * 64);
```
luminousmining: magic number

sources/algo/kheavyhash/tests/cuda_device_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+        kheavyhash::kat::POW_KAT_TIMESTAMP,
+        kheavyhash::kat::POW_KAT_NONCE,
+        out);
+    EXPECT_EQ(toArr(out), kheavyhash::kat::POW_KAT_EXPECTED);
+}
+
+
+TEST(CudaDeviceKat, KHeavyHashMatchesReference)
+{
+    uint8_t out[32];
```
luminousmining: magic number

sources/algo/kheavyhash/tests/cuda_device_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+
+TEST(CudaDeviceKat, FullPipelineMatchesReference)
+{
+    // The matrix is generated host-side (CPU reference); the CUDA device funcs do
+    // powHash -> heavyHash, mirroring the on-GPU per-nonce path.
+    kheavyhash::Matrix const    matrix{ kheavyhash::generateMatrix(kheavyhash::kat::FP_PRE) };
+    std::vector<uint16_t> const flat{ flatten(matrix) };
+
+    uint8_t h1[32];
```
luminousmining: magic number

sources/algo/kheavyhash/tests/cuda_device_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+            }
+        }
+        return f;
+    }
+}
+
+
+TEST(CudaDeviceKat, PowHashMatchesReference)
+{
+    uint8_t out[32];
```
luminousmining: magic number

sources/algo/kheavyhash/tests/cuda_device_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+        }
+        return a;
+    }
+
+
+    std::vector<uint16_t> flatten(kheavyhash::Matrix const& m)
+    {
+        std::vector<uint16_t> f;
+        f.reserve(64 * 64);
+        for (int r{ 0 }; r < 64; ++r)
```
luminousmining: magic number

sources/algo/kheavyhash/tests/cuda_device_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include <gtest/gtest.h>
+
+#include <algo/kheavyhash/cuda/kheavyhash_device.cuh>
+#include <algo/kheavyhash/matrix.hpp>
+
+namespace
+{
+    std::array<uint8_t, 32> toArr(uint8_t const* p)
+    {
+        std::array<uint8_t, 32> a{};
```
luminousmining: magic number

sources/stratum/kheavyhash.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+void stratum::StratumKHeavyHash::onUnknownMethod(boost::json::object const& root)
+{
+    std::string const method{ common::boostGetString(root, "method") };
+
+    if ("mining.authorize" == method)
+    {
+        onResponse(root);
+    }
+    // NiceHash delivers the extranonce as the subscribe reply with the bare
```
luminousmining: Move this comment in the scope `else if`

sources/algo/kheavyhash/matrix.cpp
yuzi-co: Done in `7f99eaa` — dropped the unnamed namespace; `randMatrixNoRankCheck` is now `static`.

sources/algo/kheavyhash/matrix.cpp
yuzi-co: Resolved in `7f99eaa` — the bare `64` is now `kheavyhash::MATRIX_DIM` (a `std::size_t` constant), so the bound is unsigned and named rather than a literal.

sources/algo/kheavyhash/matrix.cpp
yuzi-co: Added a bound in `7f99eaa` — `maxAttempts{ 1024 }`. A random nibble matrix is full-rank with overwhelming probability (converges in 1–2 draws), so the cap only guards a pathological stream; the loop can no longer spin forever.

documentation/BENCHMARK.md
yuzi-co: Fixed in `9e9373d` — moved the per-kernel methodology paragraph under the `### kHeavyHash` subsection, leaving the `## … RX 9070 XT` header device-level like the NVIDIA section.

sources/algo/kheavyhash/cuda/kheavyhash_device.cuh
yuzi-co: Held this one (and the `compare.cuh` request below) pending your call, because they're coupled and they remove the no-GPU test path.

The `#if defined(__CUDACC__)` shim is deliberate: with `__CUDACC__` undefined the `__device__` functions compile as plain host functions, which is what lets `tests/cuda_device_test.cpp` verify the CUDA port bit-for-bit **on a machine with no NVIDIA GPU** (it's the only thing that does). Going strictly-CUDA makes the file nvcc-only and that host KAT can no longer run.

I did apply the in-file nits that don't depend on the shim (`__restrict__`, `uint64_t* const`, explicit nibble) in `119d6e5`.

If you want the strict-CUDA form regardless, two ways to keep coverage:
1. convert `cuda_device_test` into a real `.cu` device test (runs on your CI/GPU, not host), or
2. delete `cuda_device_test` as redundant — the CPU KAT (`kheavyhash_test.cpp`) and the OpenCL KAT already gate the algorithm bit-for-bit.

Which do you prefer? I'll do the macro swap + `compare.cuh` extraction together with whichever test path you pick.

sources/algo/kheavyhash/cuda/kheavyhash_device.cuh
yuzi-co: Done in `119d6e5` — `keccakF1600(uint64_t* const a)`.

sources/algo/kheavyhash/cuda/kheavyhash_device.cuh
yuzi-co: Added in `119d6e5` — `meetsTarget` pointer params are now `__restrict__`.

sources/algo/kheavyhash/cuda/kheavyhash_device.cuh
yuzi-co: Added in `119d6e5` — `heavyHash` pointer params are now `__restrict__`.

sources/algo/kheavyhash/cuda/kheavyhash_device.cuh
yuzi-co: Done in `119d6e5` — split into named `highNibble`/`lowNibble` then packed. Behaviour-preserving: each `sum >> 10` is ≤ 14 (max accumulator 64·15·15 = 14400, `>>10` = 14), so the added `& 0xF` is a no-op; the KAT stays green.

sources/algo/kheavyhash/cuda/kheavyhash_device.cuh
yuzi-co: Deferred together with the `#if __CUDACC__` macro thread above — extracting `meetsTarget` into a `__device__` `compare.cuh` template makes it nvcc-only, which is the same strict-CUDA trade-off. I'll do it with whichever test path you choose there.

sources/algo/kheavyhash/cuda/memory.cuh
yuzi-co: Done in `7f99eaa` — `algo::kheavyhash::MATRIX_DIM * algo::kheavyhash::MATRIX_DIM` (new constant in `result.hpp`).

sources/algo/kheavyhash/opencl/tests/opencl_kat_test.cpp
yuzi-co: Done in `5ccfd60` — removed the unnamed namespace; the helpers are now `static` free functions (also avoids an ODR clash on the `flatten` name shared with `cuda_device_test.cpp`).

sources/algo/kheavyhash/opencl/CMakeLists.txt
yuzi-co: Removed in `9e9373d`.

sources/algo/kheavyhash/opencl/CMakeLists.txt
yuzi-co: Removed in `9e9373d`.

sources/algo/kheavyhash/tests/CMakeLists.txt
yuzi-co: Removed in `9e9373d`.

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
yuzi-co: Done in `c692993` — reindented; the file was rewritten and the same cleanup applied to all six `kHeavyHash_lm0..5.cl`.

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
yuzi-co: Done in `c692993` — now `rol_u64` from `kernel/common/rotate_byte.cl`.

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
yuzi-co: Done in `c692993` — moved to a new `kernel/common/load_store_le.cl` as `load_le_u64` (wired into the common deploy list).

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
yuzi-co: Done in `c692993` — moved to `kernel/common/load_store_le.cl` as `store_le_u256`.

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
yuzi-co: Done in `c692993` — the plain permutation now loops over the shared `keccak_f1600_round`. The variants whose keccak *is* the optimisation (lm3 register-resident, lm4/lm5 midstate + round-0 hoist) keep their specialised version; only the rotate was swapped to `rol_u64`.

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
yuzi-co: Done in `c692993` — dropped the hand-written `Result` struct; the kernels use the shared `t_result` from `kernel/common/result.cl`.

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
yuzi-co: Done in `c692993` — now `t_result` from `kernel/common/result.cl`, single nonce.

sources/benchmark/opencl/kheavyhash/kHeavyHash_lm0.cl
yuzi-co: Done in `c692993` — applied the dedup to all six `lm0..5`; also removed the now-dead duplicate base `powHash`/`kHeavyHash`/`heavyHash` carried by the optimised variants.

sources/algo/kheavyhash/cuda/memory.cuh
yuzi-co: Done in `7f99eaa` — `MATRIX_DIM * MATRIX_DIM`.

sources/algo/kheavyhash/cuda/memory.cuh
yuzi-co: Done in `7f99eaa` — named `RESULT_BUFFER_COUNT` (one device-side `Result` per in-flight stream — double buffering). Note: the same `sizeof(Result) * 2u` idiom is inlined in ethash/progpow/blake3/autolykos `memory.cuh` too; happy to centralise it across all algos in a follow-up, kept it kHeavyHash-scoped here.

sources/algo/kheavyhash/opencl/tests/opencl_kat_test.cpp
yuzi-co: Done in `5ccfd60` — construct `std::ifstream const file{ cand }` then test `file.good()`, under the `nullptr` guard.

sources/algo/kheavyhash/opencl/tests/opencl_kat_test.cpp
yuzi-co: Done in `5ccfd60` — `powHashMatchesReference` (all gtest case names lower-cased across both test files).

sources/algo/kheavyhash/opencl/tests/opencl_kat_test.cpp
yuzi-co: Done in `5ccfd60` — used `kheavyhash::HASH_SIZE` (new constant in `types.hpp`, kept dependency-free). Same intent as `HASH_256_LEN`, scoped to the standalone reference namespace.

sources/algo/kheavyhash/opencl/tests/opencl_kat_test.cpp
yuzi-co: Done in `5ccfd60` — `count{ 0u }`.

sources/algo/kheavyhash/opencl/tests/opencl_kat_test.cpp
yuzi-co: Done in `5ccfd60` — `nonces[...]{ 0ull, ... }`, and the array length switched to `algo::kheavyhash::MAX_RESULT`.

sources/algo/kheavyhash/tests/cuda_device_test.cpp
yuzi-co: Done in `5ccfd60` — case names lower-cased.

sources/algo/kheavyhash/tests/cuda_device_test.cpp
yuzi-co: Done in `5ccfd60` — `MATRIX_DIM * MATRIX_DIM`.

sources/algo/kheavyhash/tests/cuda_device_test.cpp
yuzi-co: Done in `5ccfd60` — `HASH_SIZE`.

sources/algo/kheavyhash/tests/cuda_device_test.cpp
yuzi-co: Done in `5ccfd60` — `HASH_SIZE` / `MATRIX_DIM`.

sources/algo/kheavyhash/tests/cuda_device_test.cpp
yuzi-co: Done in `5ccfd60` — `HASH_SIZE`.

sources/algo/kheavyhash/tests/cuda_device_test.cpp
yuzi-co: Done in `5ccfd60` — `MATRIX_DIM`.

sources/algo/kheavyhash/tests/cuda_device_test.cpp
yuzi-co: Done in `5ccfd60` — `HASH_SIZE`.

sources/stratum/kheavyhash.cpp
yuzi-co: Done in `9e9373d` — moved the rationale inside the `else if` block.

sources/algo/kheavyhash/cuda/kheavyhash_device.cuh
luminousmining: We are in a `.cuh` file, so it is intended exclusively for `cuda`.

yuzi-co: Done in `5c0922c` — strict CUDA. Dropped the `#if defined(__CUDACC__)` shim and the `KH_FN`/`KH_CONST` macros; the device functions are now plain `__device__ __forceinline__` and the constants `__device__ __constant__`, matching `crypto/cuda/keccak_f1600.cuh`.

The host-compile path is gone with it: `tests/cuda_device_test.cpp` (the only host consumer of the `.cuh`) is removed. No KAT coverage is lost — the CPU reference is the oracle, and `hashers_test`/`kheavyhash_test` already pin the identical pow / heavy / full-pipeline / target vectors, same as every other algo whose CUDA device header has no host test.

sources/algo/kheavyhash/cuda/kheavyhash_device.cuh
yuzi-co: Done in `5c0922c`. Added `isLowerLittleEndian<typename T, uint32_t N>(T const* __restrict__ r, T const* __restrict__ l)` to `sources/common/cuda/compare.cuh` — the little-endian sibling of the existing `isLowerOrEqual` (MSB at the highest index, so it scans `N-1 → 0`). The device `meetsTarget` is removed; `search.cuh` now calls `isLowerLittleEndian<uint8_t, algo::LEN_HASH_256>(pow, tgt)`.

sources/common/cuda/compare.cuh
```diff
@@ -23,3 +23,24 @@ bool isLowerOrEqual(
...
+template<typename T, uint32_t N>
+__forceinline__ __device__
+bool isLowerLittleEndian(
+    T const* __restrict__ r,
+    T const* __restrict__ l)
+{
+    #pragma unroll
+    for (int32_t i{ static_cast<int32_t>(N) - 1 }; 0 <= i; --i)
+    {
+        if (r[i] != l[i])
```
luminousmining: 2 compare `!=` and `<` keep only `<`

yuzi-co: Done in `a8f9172` — dropped the `!=`; the loop now uses only `<` (`if (r[i] < l[i]) return true; if (l[i] < r[i]) return false;`), matching the `isLowerOrEqual` sibling. Same `r <= l` semantics — the first differing element decides, all-equal returns true. Also rebased the branch onto current `main`.

PR #155, opened 2026-06-09
NetworkTCPClient: serialize concurrent writes and tie handler lifetime to the object (#139)

yuzi-co:
Closes #139 (follow-up to #132).

## Problem
`NetworkTCPClient` had three concurrency/lifetime defects on its send path:

1. **Overlapping writes.** `send()` held `txMutex` only while *launching* `async_write`, which returns immediately. Device threads (share submits) and the io thread (responses) could start two `async_write` ops on the same socket at once — Asio forbids this, and overlapping writes interleave and corrupt an outgoing Stratum frame.
2. **Handler captured `this` by raw pointer.** `onSend` dereferenced a raw `this`; if the client were destroyed before the handler ran, that's a use-after-free.
3. **`ssl::stream` read/write race.** Even with writes serialized, `async_write` was initiated from device threads while the io thread held a pending `async_read` on the same `ssl::stream` — concurrent access to one SSL state machine (UB) on secured connections.

## Changes
- **`network::WritePump`** (new, unit-tested): a small single-writer state machine — FIFO queue + one-in-flight token, mutex-guarded, `transmit` invoked outside the lock. Guarantees only one `async_write` is ever in flight. `send()` enqueues; `onSend()` drives the next.
- **`enable_shared_from_this`**: the write completion handler now captures `shared_from_this()` + the payload, so the client and buffer outlive any in-flight write. Stratum ownership moves from raw pointers to `shared_ptr` (`NewStratum`, `DeviceManager::stratums`, `stratumSmartMining`, `StratumSmartMining::stratumPool`); observers (`Device::stratum`, etc.) stay non-owning raw.
- **Strand**: a `boost::asio::strand` over `ioContext` now carries the read completion handler and every write initiation (`dispatch`), so the `ssl::stream` is only ever touched by one executor at a time. Pump = one-write-at-a-time; strand = thread-safety (both are needed).
- **SmartMining**: the child `stratumPool` shares the parent's pump, keeping a single writer on the shared socket.
- Removes the unused lockfree `tx` queue and the now-redundant `txMutex`.

## Tests
New `WritePump` gtest coverage (`sources/network/tests/`): FIFO order, single-in-flight, drain, idle/restart, error-clears-queue, and a multi-threaded enqueue/complete stress test.

## Verification
- `unit_test` green (incl. all 6 `WritePump.*`).
- Live, RX 9070 XT:
  - **zpool kawpow (plain TCP):** 24 accepted shares, concurrent device-thread submits + io-thread jobs, no frame corruption.
  - **NiceHash ethash/kawpow (plain TCP):** 526 concurrent submit frames cleanly parsed by the pool, no corruption/disconnect (the "above target" rejections there were an unrelated, pre-existing RDNA4 DAG-generation bug, fixed separately).
  - **zpool kawpow over TLS (`:51325 --ssl`):** handshake + interleaved jobs/submits, 3/3 shares accepted, 0 TLS/corruption signals — exercises the concurrent encrypted read+write path the strand fixes.

## Notes
- Receive-path remains single-reader; no behavior change to the protocol parsers.


sources/device/device_manager.hpp
```diff
@@ -70,16 +72,16 @@ namespace device
...
-                         common::Dashboard&                         board,
-                         device::Device* const                      device,
-                         double const                               hashrate,
-                         std::string const&                         host,
-                         statistical::Statistical::ShareInfo const& shareInfo);
+        void updateDevice(uint32_t const stratumUUID, bool const updateMemory, bool const updateConstants);
+        bool containStratum(uint32_t const deviceId) const;
+        std::shared_ptr<stratum::Stratum> getOrCreateStratum(algo::ALGORITHM const algorithm, uint32_t const deviceId);
+        void                              loopStatistical();
+        void                              showMiningStats(
```
luminousmining: Wrong parameter alignment.
```cpp
void showMiningStats(common::Dashboard&  board,
                     device::Device*     device);
```

sources/network/network.cpp
```diff
@@ -342,34 +354,59 @@ bool network::NetworkTCPClient::handshake()
...
+        [self, payload]()
+        {
+            if (nullptr == self->socketTCP) [[unlikely]]
+            {
+                logErr() << "Cannot send packet, socketTCP is nullptr!";
+                self->pump->onComplete(false);
+                return;
+            }
+
+            auto handler{ boost::asio::bind_executor(
```
luminousmining, line 392: Instead of:

```cpp
            auto handler{ boost::asio::bind_executor(
                self->strand,
                [self, payload](boost_error_code const& ec, std::size_t bytes)
                {
                    self->onSend(ec, bytes);
                }) };
```

I prefer:

```cpp
    auto handler
    {
        boost::asio::bind_executor
        (
            self->strand,
            [self, payload](boost_error_code const& ec, std::size_t bytes)
            {
                self->onSend(ec, bytes);
            }
        )
    };
```

Please apply this formatting consistently across the entire PR.

sources/stratum/stratums.cpp
```diff
@@ -4,71 +4,71 @@
...
         case algo::ALGORITHM::ETHASH:
         {
-            stratum = NEW(stratum::StratumEthash);
+            stratum = std::make_shared<stratum::StratumEthash>();
             break;
         }
         case algo::ALGORITHM::ETCHASH:
         {
-            stratum = NEW(stratum::StratumEtchash);
+            stratum = std::make_shared<stratum::StratumEtchash>();
```
luminousmining: We need to create new define `NEW_SHARED` -> we have tools to track the allocations

sources/network/network.cpp
yuzi-co: The established `.cpp`/`.hpp` convention here is attached-paren — `.clang-format` sets `AlignAfterOpenBracket: AlwaysBreak` with `BinPackArguments/Parameters: false`, which keeps `(` on the call line and breaks after it. Grepping all of `sources/` for a lone `(` on its own line in `.cpp`/`.hpp` returns zero matches; that style currently only appears in the CUDA `.cu`/`.cuh` files.

Two concerns with adopting paren-on-its-own-line here:

1. **It can't be expressed in our `.clang-format`** — clang-format has no option to put `(` on its own line for calls, so these spots would have to be hand-maintained against `--target format` and would drift over time.
2. **"Consistently across the entire PR" pulls in out-of-scope code** — to be genuinely consistent it would require reformatting unrelated wrapped calls this PR happens to touch (e.g. the existing `setCallbackShareStatus(std::bind(...))` blocks), which are identical to untouched copies elsewhere in the file. That expands a write-serialization bugfix into a file-wide restyle.

The PR keeps the existing attached-paren format. If paren-on-its-own-line should become the house `.cpp` convention, it can be done as a dedicated, tree-wide formatting change (with the matching `.clang-format`/tooling update) so it lands in one consistent pass rather than piecemeal here.

sources/device/device_manager.hpp
yuzi-co: Done in `24bbd59` — dropped the declaration-column alignment for the block and wrapped the parameters under the open paren.

sources/stratum/stratums.cpp
yuzi-co: Added `NEW_SHARED(type, ...)` in `custom.hpp` (mirrors `NEW`/`NEW_ARRAY`) and routed the `make_shared` call sites through it — the 13 in `stratums.cpp` plus SmartMining and the network write pump/payload. Done in `d0374a8`.

sources/network/network_callback.cpp
```diff
@@ -116,4 +122,8 @@ void network::NetworkTCPClient::onSend(boost_error_code const& ec, [[maybe_unuse
     {
         logErr() << "Fail on send message to " << host << ":" << port;
     }
+    if (nullptr != pump)
+    {
+        pump->onComplete(boost::system::errc::errc_t::success == ec);
+    }
 }
```
luminousmining: Add new line (empty) please.
A file must finish with empty line

sources/network/write_pump.cpp
```diff
@@ -0,0 +1,50 @@
...
+    }
+    transmit(next);
+}
+
+
+void network::WritePump::onComplete(bool const success)
+{
+    Payload next{};
+    {
+        std::lock_guard<std::mutex> const lock{ mutex };
```
luminousmining: Prefere `UNIQUE_LOCK`

sources/network/write_pump.hpp
```diff
@@ -0,0 +1,39 @@
...
+        void enqueue(Payload payload);
+
+        // Owner reports the in-flight write finished.
+        //   success: drop the completed head; send next if any, else go idle.
+        //   failure: drop all queued payloads (stale; connection going down) and
+        //            go idle.
+        void onComplete(bool success);
+
+      private:
+        std::mutex          mutex{};
```
luminousmining: prefere use boost mutex instead of STL

sources/network/tests/write_pump.cpp
```diff
@@ -0,0 +1,168 @@
...
+    // Stress enqueue from many threads while a single consumer completes once
+    // per transmit. The pump must never stall (every enqueue eventually
+    // transmits exactly once).
+    std::atomic<int>   transmitted{ 0 };
+    network::WritePump pump{ [&transmitted](network::WritePump::Payload const&)
+                             {
+                                 transmitted.fetch_add(1, std::memory_order_relaxed);
+                             } };
+
+    constexpr int producers{ 8 };
```
luminousmining: constexprt must be in MAJ or replace by `int const`

sources/network/network_callback.cpp
yuzi-co: Done in `1700ecf` — added the terminating empty newline.

sources/network/write_pump.cpp
yuzi-co: Done in `1700ecf` — both lock sites (`enqueue` and `onComplete`) now use `UNIQUE_LOCK(mutex)`, with `#include <common/custom.hpp>` for the macro.

sources/network/write_pump.hpp
yuzi-co: Done in `1700ecf` — the member is now `boost::mutex` (include swapped `<mutex>` → `<boost/thread.hpp>`). This also pairs with the `UNIQUE_LOCK` change above, which is `boost::unique_lock<boost::mutex>`.

sources/network/tests/write_pump.cpp
yuzi-co: Done in `1700ecf` — replaced both with `int const`. Verified: cross-built `unit_test` (`GPU=amd`) and `WritePump.*` pass 6/6 on the host, including `ConcurrentEnqueueAndCompleteNeverStalls`.

PR #157, opened 2026-06-09
fix(network): verify TLS with the Windows chain engine on Windows

yuzi-co:
## Problem

On Windows the miner seeded OpenSSL only with the `ROOT` store. OpenSSL has no AIA fetching, so a pool that serves an **incomplete chain** (leaf only, omitting the intermediate) fails verification even though Windows trusts the certificate — a very common pool misconfiguration. Concrete case: `pool.woolypooly.com:3106` (TLS-only) ships the leaf without the Sectigo R36 intermediate, so `--ssl=true` failed with `TLS certificate chain verification failed` / `Cannot connect to pool with SSL option`.

## Fix

Hand the whole-chain trust decision to the native Windows chain engine, which supplies the system roots, AIA-fetches a missing intermediate, and validates the hostname — the same path Schannel/WinHTTP use.

- **`doSecureConnection`**: no longer seeds an OpenSSL trust store on Windows. `verify_peer` stays set so the callback still runs; the empty OpenSSL store just makes `preverified` always false (expected). Linux path unchanged (default paths + `ca-certificates.crt` + `SSL_CERT_FILE`).
- **`onVerifySSL`**: on Windows, defer to the leaf (depth 0) and call the new `verifyChainWithWindows()`; let higher chain levels pass so OpenSSL walks down to the leaf. Other platforms keep the OpenSSL `preverified` + `host_name_verification` path.
- **`verifyChainWithWindows`**: DER-encodes the peer-sent certs into an in-memory `HCERTSTORE`, calls `CertGetCertificateChain` (default engine = system roots + AIA fetch; no revocation flags = no CRL/OCSP network dependency), then `CertVerifyCertificateChainPolicy(CERT_CHAIN_POLICY_SSL)` to validate the chain **and** the hostname. Requires the `serverAuth` EKU. Frees the chain/contexts/store on every path.
- Moved the `crypt32` pragma + `wincrypt.h` include into `network_callback.cpp` (where they are now used) and dropped them from `network.cpp`.

More robust than seeding the cached `CA` store: it does not depend on the intermediate already being cached, so it works on a fresh machine and on the first connection.

## Security

Strictly tightens verification. Depth-0 delegation is safe — OpenSSL always invokes the callback at depth 0 (leaf) last, so the authoritative decision is made there; untrusted-root / self-signed / expired / name-mismatch all surface as a non-zero `dwError` and are rejected.

## Validation

- Compiles under the clang-cl cross-toolchain (`docker-build.ps1 -Os windows-cross -Gpu amd`).
- **Live-verified**: connecting with `--ssl=true` to `pool.woolypooly.com:3106` now reaches `Stratum connected!` and exchanges data over TLS, where the `ROOT`-only build failed.

## Notes

- No revocation checking (deliberate, avoids a CRL/OCSP network dependency; matches prior behavior).
- First handshake to an incomplete-chain pool may briefly block while Windows AIA-fetches the intermediate (then cached).

### DIFF
--- a/sources/network/network_callback.cpp
+++ b/sources/network/network_callback.cpp
@@ -3,6 +3,131 @@
 #include <common/log/log.hpp>
 #include <network/network.hpp>
 
+#if defined(_WIN32)
+#pragma comment(lib, "crypt32.lib")
+#include <string>
+
+#include <wincrypt.h>
+
+
+namespace
+{
+    // Validate the peer-presented certificate chain with the native Windows chain engine.
+    // Unlike OpenSSL, this supplies the system trust anchors, fetches any intermediate the
+    // server omitted via the certificate's AIA extension, and matches the hostname. Returns
+    // true only when the chain is trusted AND valid for `host`.
+    bool verifyChainWithWindows(X509_STORE_CTX* ctx, std::string const& host)
+    {
+        STACK_OF(X509)* const peerCerts{ X509_STORE_CTX_get0_untrusted(ctx) };
+        X509* const           leaf{ X509_STORE_CTX_get0_cert(ctx) };
+        if (nullptr == leaf)
+        {
+            logErr() << "TLS certificate is missing.";
+            return false;
+        }
+
+        // In-memory store holding the certs the server actually sent, so the chain engine can
+        // use any intermediates it did provide (and AIA-fetch the ones it did not).
+        HCERTSTORE const winStore{
+            CertOpenStore(CERT_STORE_PROV_MEMORY, X509_ASN_ENCODING, 0, CERT_STORE_CREATE_NEW_FLAG, nullptr)
+        };
+        if (nullptr == winStore)
+        {
+            logErr() << "Cannot allocate the Windows certificate store.";
+            return false;
+        }
+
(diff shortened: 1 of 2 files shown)


sources/network/network_callback.cpp
```diff
@@ -35,6 +160,18 @@ void network::NetworkTCPClient::asyncReceive()
 
 bool network::NetworkTCPClient::onVerifySSL(bool preverified, boost_verify_context& ctx)
 {
+#if defined(_WIN32)
+    // OpenSSL has no trust anchors on Windows (see doSecureConnection), so `preverified`
+    // is always false here. The real trust decision is delegated to the Windows chain
+    // engine, made once at the leaf (depth 0); let the higher chain levels pass so OpenSSL
+    // keeps walking down to the leaf.
+    (void) preverified;
```
luminousmining: use attribut [[maybe_unsued]] instead of this cast

yuzi-co: Done in `e93887b` — replaced the `(void) preverified;` cast with `[[maybe_unused]]` on the parameter. The attribute is the right fit here since `preverified` is genuinely used on the non-Windows `#else` path and only unused on the Windows chain-engine path.

PR #160, opened 2026-06-09
fix(autolykos): mine and submit accepted shares on NiceHash (boundaryU64, DAG overflow, hostHeader)

yuzi-co:
## Summary
Autolykos2 on NiceHash never started hashing, and once that was fixed it never landed an accepted share. This fixes the full chain end-to-end. **AMD is live-verified** on an RX 9070 XT against `autolykos.eu.nicehash.com`: **3 accepted shares, 0 rejected, 0 "above target"**. The CUDA path gets the analogous fix but is **untested** (see Caveats).

## Root causes fixed
1. **Jobs never dispatched.** NiceHash sends no `mining.set_difficulty`; the per-job target is embedded in `mining.notify` params[6]. `onMiningNotify` set `jobInfo.boundary` but not `jobInfo.boundaryU64`, so `isValidJob()` dropped every job (`Boundary U64 == 0ull`). Derive `boundaryU64` from the notify target.
2. **"Share above target" false positives.** A 32-bit index overflow in the AMD DAG kernels -- `dag.cl` `tid*32+31` (last-byte zeroing) and `search.cl` `h3*32+i` (lookup) -- wraps for index >= 2^27 (the 4 GiB boundary). On a live table (`N ~ 216M`, ~6.9 GB) this left every element above 4 GiB with a non-zero top byte and made search read the wrong element, corrupting the verify sum. Cast both indices to 64-bit.
3. **Every real share dropped before submit.** `parameters.hostHeader` was never assigned in the AMD resolver (the NVIDIA resolver sets it), so the CPU re-check in `getResultCache` (`isValidShare`) hashed candidates against a zero header and rejected them all. Set it in `updateConstants`.
4. **CUDA parity (untested).** The same overflow exists in `cuda/dag.cuh` and two benchmark prehash kernels; fixed with the same cast. Tracked in #159.

## Changes
- `stratum/autolykos_v2.cpp` -- derive `boundaryU64` from the notify-embedded target.
- `algo/autolykos/opencl/{dag,search}.cl` -- 64-bit DAG indexing.
- `resolver/amd/autolykos_v2.cpp` -- propagate `hostHeader` in `updateConstants`.
- `algo/autolykos/cuda/dag.cuh`, `benchmark/cuda/autolykos_v2/*prehash*.cu` -- 64-bit indexing (untested on NVIDIA).
- Tests: stratum notify (`boundaryU64 != 0`), CPU `isValidShare` canonical Ergo vector, AMD GPU canonical vector (raw verify buffer), AMD DAG full-coverage across the 4 GiB boundary at a live height, and a `findNonce` guard that `updateConstants` propagates `hostHeader`.

## Testing
- 7 autolykos unit tests green (cross-built, run on RX 9070 XT / gfx1201).
- Live: NiceHash autolykos -- 3 accepted, 0 rejected.

## Caveats / follow-ups
- **CUDA fixes unverified** (no NVIDIA hardware) -- see #159.
- `format-check` not run locally (no clang-format available); manual review clean.
- Pre-existing, out of scope: the iGPU as `Device[1]` crashes the process on autolykos (workaround: `--devices_disable 1`).

### DIFF
--- a/sources/resolver/amd/autolykos_v2.cpp
+++ b/sources/resolver/amd/autolykos_v2.cpp
@@ -30,9 +30,7 @@ resolver::ResolverAmdAutolykosV2::~ResolverAmdAutolykosV2()
 bool resolver::ResolverAmdAutolykosV2::updateMemory(stratum::StratumJobInfo const& jobInfo)
 {
     ////////////////////////////////////////////////////////////////////////////
-    if (   nullptr == clContext
-        || nullptr == clQueue[0]
-        || nullptr == clQueue[1]) [[unlikely]]
+    if (nullptr == clContext || nullptr == clQueue[0] || nullptr == clQueue[1]) [[unlikely]]
     {
         return false;
     }
--- a/sources/resolver/amd/autolykos_v2.cpp
+++ b/sources/resolver/amd/autolykos_v2.cpp
@@ -114,11 +114,6 @@ bool resolver::ResolverAmdAutolykosV2::updateConstants(stratum::StratumJobInfo c
         return false;
     }
 
-    ////////////////////////////////////////////////////////////////////////////
-    // getResultCache() re-checks GPU candidates on the CPU via isValidShare(), which
-    // reads parameters.hostHeader. It must track the job header (the NVIDIA resolver
-    // does the same in updateMemory); without this it stayed zero-initialised, so the
-    // CPU re-check hashed the wrong header and rejected every valid GPU share.
     parameters.hostHeader = jobInfo.headerHash;
 
     uint32_t const* const header{ jobInfo.headerHash.word32 };
--- a/sources/resolver/amd/autolykos_v2.cpp
+++ b/sources/resolver/amd/autolykos_v2.cpp
@@ -117,6 +117,12 @@ bool resolver::ResolverAmdAutolykosV2::updateConstants(stratum::StratumJobInfo c
     }
 
     ////////////////////////////////////////////////////////////////////////////
+    // getResultCache() re-checks GPU candidates on the CPU via isValidShare(), which
+    // reads parameters.hostHeader. It must track the job header (the NVIDIA resolver
+    // does the same in updateMemory); without this it stayed zero-initialised, so the
+    // CPU re-check hashed the wrong header and rejected every valid GPU share.
+    parameters.hostHeader = jobInfo.headerHash;
+
     uint32_t const* const header{ jobInfo.headerHash.word32 };
     if (false == parameters.headerCache.setBufferDevice(clQueue[currentIndexStream], header))
     {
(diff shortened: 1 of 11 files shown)


sources/resolver/amd/autolykos_v2.cpp
```diff
@@ -117,6 +115,12 @@ bool resolver::ResolverAmdAutolykosV2::updateConstants(stratum::StratumJobInfo c
     }
 
     ////////////////////////////////////////////////////////////////////////////
+    // getResultCache() re-checks GPU candidates on the CPU via isValidShare(), which
```
luminousmining: delete comment

sources/stratum/tests/autolykos_v2_notify.cpp
```diff
@@ -0,0 +1,101 @@
...
+//
+// The fix recomputes boundaryU64 from the notify boundary, mirroring the other
+// stratums (ethash/progpow/blake3) whose notify handlers always do so.
+//
+// No network needed: updateJob() is overridden to a no-op so onMiningNotify()'s
+// only observable effect is the mutated jobInfo.
+////////////////////////////////////////////////////////////////////////////////
+
+
+namespace
```
luminousmining: no empty namespace

sources/stratum/tests/autolykos_v2_notify.cpp
```diff
@@ -0,0 +1,101 @@
...
+        };
+        return root;
+    }
+}
+
+
+struct StratumAutolykosV2NotifyTest : public testing::Test
+{
+    StratumAutolykosV2NotifyTest() = default;
+    ~StratumAutolykosV2NotifyTest() override = default;
```
luminousmining: override ?

sources/stratum/autolykos_v2.cpp
```diff
@@ -78,6 +78,10 @@ void stratum::StratumAutolykosV2::onMiningNotify(boost::json::object const& root
     jobInfo.headerHash = algo::toHash256(params.at(2).as_string().c_str());
     jobInfo.boundary = algo::toHash2<algo::hash256, algo::hash512>(
         algo::toLittleEndian<algo::hash512>(algo::decimalToHash<algo::hash512>(params.at(6).as_string().c_str())));
+    // NiceHash sends no mining.set_difficulty/set_target -- the per-job target is
```
luminousmining: delete comment

sources/resolver/amd/autolykos_v2.cpp
yuzi-co: Done in 6163198 — removed the explainer comment, kept `parameters.hostHeader = jobInfo.headerHash;`.

sources/stratum/tests/autolykos_v2_notify.cpp
yuzi-co: Done in 6163198 — dropped the anonymous namespace; helpers now at file scope (names are unique across test TUs).

sources/stratum/tests/autolykos_v2_notify.cpp
yuzi-co: Done in 6163198 — dropped `override` on the dtor to match the dominant test convention (`~XxxTest() = default;`).

sources/stratum/autolykos_v2.cpp
yuzi-co: Done in 6163198 — removed the NiceHash explainer comment above `boundaryU64`.

PR #161, opened 2026-06-09
feat(blake3): Alephium AMD/OpenCL mining on a shared BLAKE3 primitive

yuzi-co:
## Summary

Adds **Alephium (BLAKE3) mining on AMD/OpenCL**, built on a new reusable BLAKE3 primitive. (BLAKE3 previously had only an unused NVIDIA-side skeleton.)

- **Shared BLAKE3 primitive** — canonical single-chunk OpenCL kernel (`algo/crypto/opencl/blake3.cl`), vendored upstream BLAKE3 C reference (`algo/crypto/reference/blake3/`), and a GPU-portable KAT verified bit-exact against the reference. Designed to be reused (Alephium here; FishHash/ethashb3 later).
- **Alephium AMD resolver + stratum** — `ResolverAmdBlake3` (non-DAG), `StratumBlake3`, with on-GPU and offline protocol tests. PoW = `BLAKE3(BLAKE3(nonce24 ‖ headerBlob302))`. The Alephium search kernel `#include`s the shared primitive (no duplication).
- **Benchmark** — DAG-free `runAmdBlake3()` benches the production Alephium `search` kernel; `BUILD_BENCH` opt-in added to the windows-cross image (default OFF).
- **Docs** — ARCHITECTURE / ADD_ALGORITHM / BENCHMARK updated to reflect BLAKE3 on AMD + NVIDIA.

## Test Plan

- [ ] `unit_test --gtest_filter=*Blake3*:*Alephium*` → 12/12 green (incl. on-GPU `ResolverBlake3AmdTest.findNonce`)
- [ ] Live-verified: accepted shares on HeroMiners (alephium :1199), RX 9070 XT
- [ ] `benchmark.exe` `[AMD] BLAKE3` ≈ 1.18 GH/s steady-state (8192×128, gfx1201)

## Notes

- Verified on AMD RX 9070 XT (gfx1201). NVIDIA path compiles but is not live-verified in this cycle.
- A BLAKE3 rotate micro-opt (OpenCL `rotate()` built-in vs hand-rolled shift-or) was A/B-benchmarked and found to be a wash on gfx1201 (the compiler already lowers shift-or to `v_alignbit`); shift-or retained.


POOLS.md
```diff
@@ -27,6 +27,7 @@ progpow-z `miner --stratum="ethproxy" --host=zano.luckypool.io --port=8866 --alg
 
 ## HEROMINERS
 kawpow `miner --host=de.quai.herominers.com --port=1185 --algo="kawpow" --wallet=YOUR_WALLET --workername="YOUR_WORKERNAME"`  
+alephium `miner --host=alephium.herominers.com --port=1199 --algo="blake3" --wallet=YOUR_BASE58_ALPH_ADDRESS --workername="YOUR_WORKERNAME"`  (plaintext; verified live ~1.71 GH/s on RX 9070 XT)
```
luminousmining: delete the part `(plaintext; verified live ~1.71 GH/s on RX 9070 XT) `

sources/algo/blake3/opencl/tests/blake3_kat.cpp
```diff
@@ -0,0 +1,213 @@
...
+            clCheck(err, "clCreateContext");
+            queue = clCreateCommandQueueWithProperties(context, device, nullptr, &err);
+            clCheck(err, "clCreateCommandQueue");
+
+            // The mining kernel #includes the shared BLAKE3 primitive. The miner's
+            // kernel generator resolves that include, but clCreateProgramWithSource
+            // does not -- so read both and prepend the primitive (dropping the
+            // mining kernel's #include line). Prefer the copies deployed next to the
+            // binary (shipped exe); fall back to the in-tree source paths (CI).
+            auto readKernel = [](char const* deployed, char const* fallback) -> std::string
```
luminousmining: using { instead of = to assign variable

sources/algo/blake3/opencl/tests/blake3_kat.cpp
```diff
@@ -0,0 +1,213 @@
...
+            std::string       mining{ readKernel("kernel/blake3/blake3.cl", BLAKE3_CL_PATH) };
+
+            std::string const include{ "#include \"kernel/crypto/blake3.cl\"" };
+            size_t const      incPos{ mining.find(include) };
+            if (std::string::npos != incPos)
+            {
+                mining.erase(incPos, include.size());
+            }
+
+            std::string const src{ crypto + "\n" + mining };
```
luminousmining: Look about ProgPow is build, you don't need your lambda `readKernel`.
We have KernelGenerator `https://github.com/luminousmining/miner/blob/main/sources/common/kernel_generator/kernel_generator.hpp` and function `addInclude`

sources/algo/blake3/opencl/tests/blake3_kat.cpp
```diff
@@ -0,0 +1,213 @@
...
+            char const*       srcPtr{ src.c_str() };
+            size_t const      srcLen{ src.size() };
+
+            program = clCreateProgramWithSource(context, 1, &srcPtr, &srcLen, &err);
+            clCheck(err, "clCreateProgramWithSource");
```
luminousmining: We have already some OPENCL checker : https://github.com/luminousmining/miner/blob/main/sources/common/error/opencl_error.hpp

sources/resolver/nvidia/blake3.cpp
```diff
@@ -137,8 +139,10 @@ void resolver::ResolverNvidiaBlake3::submit(stratum::Stratum* const stratum)
         {
             for (uint32_t i{ 0u }; i < resultShare.count; ++i)
             {
+                // Fixed 16-hex-char (8-byte) field first — std::hex alone drops leading
```
luminousmining: delete this comments

sources/algo/crypto/opencl/CMakeLists.txt
```diff
@@ -28,8 +29,11 @@ add_custom_target(copy_crypto_opencl_files ALL
...
+    ${OUT_CRYPTO}/blake3.cl
     ${OUT_CRYPTO}/fnv1.cl
     ${OUT_CRYPTO}/keccak_f1600.cl
     ${OUT_CRYPTO}/keccak_f800.cl
     ${OUT_CRYPTO}/kiss99.cl
-)
\ No newline at end of file
+)
+
+add_subdirectory(tests)
```
luminousmining: add new line (empty line) please

POOLS.md
yuzi-co: Done in `dfbd09c` — dropped the parenthetical.

sources/algo/crypto/opencl/CMakeLists.txt
yuzi-co: Done in `dfbd09c` — added the terminating newline.

sources/resolver/nvidia/blake3.cpp
yuzi-co: Condensed to a single line in `dfbd09c`. I kept the *why* (the pool recomputes the 24-byte nonce, so the leading zeros must be preserved) rather than deleting outright, since it is not obvious from the `setw(16) << setfill(0)` alone — without it the zero-padding is easy to "simplify" away and silently misalign the nonce. Happy to drop it entirely if you would prefer.

sources/algo/blake3/opencl/tests/blake3_kat.cpp
yuzi-co: Agreed the hand-rolled `readKernel` + `#include`-stripping is reinventing what `KernelGenerator::addInclude` does. Id like to do this as a focused follow-up rather than in this PR, for three reasons:

1. Its a harness rewrite, not a snippet swap: `KernelGenerator::build()` takes `cl::Context*`/`cl::Device*`, but this KAT is raw C (`clCreateProgramWithSource`), so `SetUpTestSuite` has to be reworked onto the `cl::` wrappers, and it builds two kernels (`test_hash` + `search`) from one program.
2. The `readKernel` fallback is load-bearing: it tries the deployed `kernel/...` path *and* an in-tree absolute fallback so the KAT runs in CI/POCL where the `kernel/` tree isnt copied next to `unit_test`. Id need to preserve that under `addInclude`.
3. The kHeavyHash OpenCL KAT uses the identical raw pattern. Id rather convert both KATs to `KernelGenerator` consistently in one change (with a GPU/POCL run to verify) than diverge here.

OK to track it as a follow-up across both KATs?

sources/algo/blake3/opencl/tests/blake3_kat.cpp
yuzi-co: This `readKernel` lambda is removed by the `KernelGenerator` conversion above, so Ill fold the `{}`-init into that follow-up rather than restyle a line thats about to go away.

sources/algo/blake3/opencl/tests/blake3_kat.cpp
yuzi-co: Id keep the gtest checker here: a unit test has to *fail* when an OpenCL call errors. `clCheck` uses `ASSERT_EQ`, so on a CL error it stops and marks the test failed. `OPENCL_ER` (the miner macro) logs the error and continues — so a broken kernel could slip through as a passing test. When we move this harness to `KernelGenerator` (the follow-up above), `OPENCL_ER` fits the program-build step, but the tests own pass/fail checks should stay gtest assertions.

sources/algo/blake3/opencl/tests/blake3_kat.cpp
```diff
@@ -0,0 +1,182 @@
...
+TEST_F(Blake3Cl, TestHashMatchesHostReference)
+{
+    try
+    {
+        algo::hash3072 const header{ algo::toHash<algo::hash3072>(HEADER_HEX, algo::HASH_SHIFT::LEFT) };
+        algo::hash256        expected{};
+        algo::blake3::hashRef(header, NONCE, expected);
+
+        uint32_t   out[8]{};
+        cl::Buffer hdrBuf{ context, CL_MEM_READ_ONLY | CL_MEM_COPY_HOST_PTR, sizeof(header),
```
luminousmining: I prefer full variable names — I assume `hdr` = `header`?

sources/algo/blake3/opencl/tests/blake3_kat.cpp
```diff
@@ -0,0 +1,182 @@
...
+                           const_cast<uint32_t*>(header.word32) };
+        cl::Buffer outBuf{ context, CL_MEM_WRITE_ONLY, sizeof(out) };
+
+        cl::Kernel k{ kernel("test_hash") };
```
luminousmining: k -> kernel

sources/algo/blake3/opencl/tests/blake3_kat.cpp
```diff
@@ -0,0 +1,182 @@
...
+        algo::blake3::hashRef(header, NONCE, digest);
+
+        uint32_t const bigIndex{ algo::blake3::chainIndex(digest) };
+        uint32_t const fromGroup{ bigIndex / 4u };
+        uint32_t const toGroup{ bigIndex % 4u };
+
+        Result     result{};
+        cl::Buffer hdrBuf{ context, CL_MEM_READ_ONLY | CL_MEM_COPY_HOST_PTR, sizeof(header),
+                           const_cast<uint32_t*>(header.word32) };
+        cl::Buffer tgtBuf{ context, CL_MEM_READ_ONLY | CL_MEM_COPY_HOST_PTR, sizeof(digest.ubytes), digest.ubytes };
```
luminousmining: tgt -> target ?

sources/algo/blake3/opencl/tests/blake3_kat.cpp
luminousmining: buf -> buffer

sources/algo/blake3/opencl/tests/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# BLAKE3_CL_PATH is set as a target-level define (not a source-file property):
+# ${UNIT_TEST_EXE} is defined in sources/CMakeLists.txt, a different directory
+# scope, so set_source_files_properties here would not reach its compile.
+if (BUILD_EXE_UNIT_TEST AND BUILD_AMD)
+    target_sources(${UNIT_TEST_EXE} PUBLIC
+        ${CMAKE_CURRENT_SOURCE_DIR}/blake3_kat.cpp
+    )
+    target_compile_definitions(${UNIT_TEST_EXE} PRIVATE
+        "BLAKE3_CL_PATH=\"${CMAKE_CURRENT_SOURCE_DIR}/../blake3.cl\""
```
luminousmining: I don't understand why you need such a workaround.
Look at the other unit tests using OpenCL — the `.cl` file is found easily because it is copied via CMakeLists. You can do the same thing with this specific file and therefore have the file in the right place.

sources/algo/blake3/opencl/blake3.cl
```diff
@@ -0,0 +1,145 @@
...
+//
+// The shared BLAKE3 primitive (sources/algo/crypto/opencl/blake3.cl) is prepended
+// by the builder, not #include'd: both the production resolver (resolver/amd/blake3.cpp)
+// and the host KAT (opencl/tests/blake3_kat.cpp) chain appendFile(crypto) + appendFile(this),
+// the same way autolykos chains its files. This keeps the assembled source self-contained
+// so it compiles without an -I/-deployed-tree dependency (the KAT runs from a CWD that has
+// no kernel/ tree). The LM_BLAKE3_CL include guard still makes prepending idempotent.
+
+#ifndef MAX_RESULT
+#define MAX_RESULT 4
```
luminousmining: This variable should be defined through the kernel generator via `addDefine`.
Take the other OpenCL kernels as an example.

sources/algo/blake3/opencl/blake3.cl
```diff
@@ -0,0 +1,145 @@
...
+#endif
+
+inline uint bswap32(uint const x)
```
luminousmining: There is a simpler approach:

```
uint result = as_uint(as_uchar4(x).s3210);
```

Since this is a "common" function, it would be good to place it in the `rotate_byte.cl` file:

```
inline
uint bswap32(uint x)
{
    uchar4 b = as_uchar4(x);
    return as_uint((uchar4)(b.s3, b.s2, b.s1, b.s0));
}
```

sources/algo/blake3/opencl/blake3.cl
```diff
@@ -0,0 +1,145 @@
...
+    compress8(vector, block2, 32u, 11u);                 // CHUNK_START|END|ROOT
+
+    for (uint i = 0u; i < 8u; ++i)
+    {
+        out8[i] = vector[i];
+    }
+}
+
+// digest <= target, byte-wise from index 0 (matches common/cuda/compare.cuh).
+inline bool isLowerOrEqual(uchar const* r, uchar const* l)
```
luminousmining: inline
bool isLowerOrEqual(uchar const* r, uchar const* l)

sources/algo/blake3/opencl/blake3.cl
```diff
@@ -0,0 +1,145 @@
+// Blake3 (Alephium) OpenCL mining kernels. The BLAKE3 compression primitive is the
```
luminousmining, line 1: Many functions are missing the `inline` prefix — is that intentional?

sources/algo/blake3/opencl/blake3.cl
```diff
@@ -0,0 +1,145 @@
...
+typedef struct __attribute__((aligned(8)))
+{
+    uchar found;
+    uint  count;
+    ulong nonces[MAX_RESULT];
+} Result;
+
+// KAT entry: digest for one nonce. header = headerBlob words (left-aligned), out = 8 u32.
+__kernel void test_hash(__global uint const* header, ulong const nonce, __global uint* out)
+{
```
luminousmining: __kernel
void test_hash(__global uint const* header, ulong const nonce, __global uint* out)

sources/algo/blake3/opencl/blake3.cl
```diff
@@ -0,0 +1,145 @@
...
+    uint h[8];
+    powHash(header, nonce, h);
+    for (uint i = 0u; i < 8u; ++i) { out[i] = h[i]; }
```
luminousmining: no one-line

sources/algo/blake3/opencl/blake3.cl
```diff
@@ -0,0 +1,145 @@
...
+{
+    uchar found;
+    uint  count;
+    ulong nonces[MAX_RESULT];
+} Result;
+
+// KAT entry: digest for one nonce. header = headerBlob words (left-aligned), out = 8 u32.
+__kernel void test_hash(__global uint const* header, ulong const nonce, __global uint* out)
+{
+    uint h[8];
```
luminousmining: h -> header

sources/algo/blake3/opencl/blake3.cl
```diff
@@ -0,0 +1,145 @@
...
+    powHash(header, nonce, h);
+    for (uint i = 0u; i < 8u; ++i) { out[i] = h[i]; }
+}
+
+// Mining kernel: nonce = startNonce + gid. Hit iff digest <= target AND the
+// chain index (digest byte 31 % 16) maps to (fromGroup, toGroup).
+__kernel void search(__global uint const*  header,
```
luminousmining: __kernel
void search

luminousmining: I prefer to have the pointers first, followed by the primitive variables.

sources/algo/blake3/opencl/blake3.cl
```diff
@@ -0,0 +1,145 @@
...
+                     __global uchar const* target,
+                     ulong const           startNonce,
+                     uint const            fromGroup,
+                     uint const            toGroup,
+                     __global Result*      result)
+{
+    ulong const nonce = startNonce + (ulong)get_global_id(0);
+
+    uint h[8];
```
luminousmining: header

sources/algo/blake3/opencl/blake3.cl
```diff
@@ -0,0 +1,145 @@
...
+    powHash(header, nonce, h);
+
+    uchar tgt[32];
+    for (uint i = 0u; i < 32u; ++i) { tgt[i] = target[i]; }
```
luminousmining: no one-line

sources/algo/blake3/blake3_pow.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include "blake3.h" // vendored reference, via blake3_ref lib
+
+#include <algo/blake3/blake3.hpp> // CHAIN_NUMBER
+#include <algo/blake3/blake3_pow.hpp>
+
+
+namespace
+{
+    inline uint32_t bswap32(uint32_t const x)
```
luminousmining: already exist from native `bitwise.hpp`.
You can call `bswap32` or `be::uint32(x)`

sources/algo/blake3/blake3_pow.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+
+void algo::blake3::hashRef(algo::hash3072 header, uint64_t const nonce, algo::hash256& out)
+{
+    // Reconstruct the 326-byte Alephium mining input exactly as the kernel does:
+    // word[0..1] = big-endian search value, word[2..5] = 0 (nonce bytes 8..23),
+    // word[6..81] = headerBlob (header.word32[0..75]). 326 bytes = 81 little-endian
+    // words + the low 2 bytes of word[81].
+    uint32_t words[96]{};
+    words[0] = bswap32(static_cast<uint32_t>(nonce >> 32));
```
luminousmining: use `cast.hpp`

sources/algo/blake3/CMakeLists.txt
```diff
@@ -2,9 +2,10 @@ file(GLOB HEADERS "*.hpp")
 file(GLOB SOURCES "*.cpp")
 
 if (BUILD_EXE_MINER)
+    # blake3_pow.cpp is the host PoW oracle used only by the unit tests; it (and the vendored
```
luminousmining: If `.cpp` files are present but not compiled, they should either be deleted or wrapped with a `UNIT_TEST` guard, similar to how `ENABLE_AMD` and `ENABLE_CUDA` are handled.

And tou can remove this comment

sources/algo/crypto/opencl/tests/blake3_kat.cl
```diff
@@ -0,0 +1,19 @@
+// Test-only wrapper kernel for the shared BLAKE3 primitive. The host harness
+// concatenates sources/algo/crypto/opencl/blake3.cl in front of this file before
+// building, so blake3_hash_chunk is in scope (no OpenCL #include needed).
+__kernel void blake3_kat(__global uchar const* in, uint const len, uint const outlen, __global uchar* out)
```
luminousmining: __kernel
void

sources/benchmark/amd/blake3.cpp
```diff
@@ -0,0 +1,140 @@
...
+    {
+        logErr() << "Fail to copy header in cache";
+    }
+    if (false == targetCache.setBufferDevice(&propertiesAmd.clQueue, &target))
+    {
+        logErr() << "Fail to copy target in cache";
+    }
+
+    ////////////////////////////////////////////////////////////////////////////
+    auto benchBlake3 =
```
luminousmining: use { instead of =

sources/benchmark/amd/blake3.cpp
```diff
@@ -0,0 +1,140 @@
...
+                    .enqueueNDRangeKernel(clKernel, cl::NullRange, cl::NDRange(globalSize), cl::NDRange(localSize)));
+            OPENCL_ER(propertiesAmd.clQueue.finish());
+            stopChrono(dashboard);
+        }
+
+        return true;
+    };
+
+    ////////////////////////////////////////////////////////////////////////////
+    auto const runKernel = [&](std::string const& name)
```
luminousmining: use { instead of =

sources/benchmark/amd/blake3.cpp
```diff
@@ -0,0 +1,140 @@
...
+        if (false == algo.isKernelEnabled(name))
+        {
+            return;
+        }
+        KernelParams const p{ algo.resolveKernel(name) };
+        benchBlake3(name, p.loop, p.threads, p.blocks);
+    };
+
+    ////////////////////////////////////////////////////////////////////////////
+    runKernel("blake3"); // Alephium production search kernel — throughput baseline
```
luminousmining: delete this comment ;)

sources/common/kernel_generator/opencl.hpp
```diff
@@ -12,13 +12,14 @@ namespace common
     struct KernelGeneratorOpenCL : common::KernelGenerator
     {
       public:
-        cl::Kernel clKernel{};
+        cl::Kernel  clKernel{};
+        // Public so a single built program can yield more than one kernel by name
+        // (the OpenCL KATs pull test_hash/search/etc. from one build). Production
+        // code only reads clKernel; widening visibility changes nothing for it.
+        cl::Program clProgram{};
```
luminousmining, line 19: I don't see `clProgram` called in your test

sources/common/mocker/stratum.hpp
```diff
@@ -22,6 +23,14 @@ namespace common
                 paramSubmit = params;
             }
 
+            // Object-form submit (e.g. Blake3/Alephium): the base logs "not implemented",
```
luminousmining: delete this comment ;)

sources/resolver/amd/blake3.cpp
```diff
@@ -0,0 +1,233 @@
...
+{
+    auto& clKernel{ kernelGenerator.clKernel };
+    OPENCL_ER(clKernel.setArg(0u, *(parameters.headerCache.getBuffer())));
+    OPENCL_ER(clKernel.setArg(1u, *(parameters.targetCache.getBuffer())));
+    OPENCL_ER(clKernel.setArg(2u, jobInfo.nonce));
+    OPENCL_ER(clKernel.setArg(3u, jobInfo.fromGroup));
+    OPENCL_ER(clKernel.setArg(4u, jobInfo.toGroup));
+    OPENCL_ER(clKernel.setArg(5u, *(parameters.resultCache.getBuffer())));
+
+    size_t const globalSize{ static_cast<size_t>(getBlocks()) * static_cast<size_t>(getThreads()) };
```
luminousmining: cast.hpp

sources/algo/blake3/opencl/blake3.cl
yuzi-co: Done in `6a6bcc9`. `MAX_RESULT` now comes solely from the kernel generator via `addDefine` — I removed the in-kernel `#ifndef MAX_RESULT / #define MAX_RESULT 4 / #endif` fallback.

It is defined at every site that builds the kernel:
- `resolver/amd/blake3.cpp` — `addDefine("MAX_RESULT", algo::blake3::MAX_RESULT)`
- `benchmark/amd/blake3.cpp` — same
- `opencl/tests/blake3_kat.cpp` — same

The generator emits the `#define` ahead of the appended kernel source, so `MAX_RESULT` is defined before `blake3.cl` uses it (the `Result` struct and the `search` bounds check) — same pattern as `progpow.cl`, which carries no in-kernel define. The value is sourced from `algo::blake3::MAX_RESULT` (`result.hpp`), so the host result buffer and the kernel stay in sync from one constant. With the fallback gone, a build site that omits the `addDefine` now fails to compile instead of silently defaulting to 4.

sources/benchmark/amd/blake3.cpp
yuzi-co: Switched benchBlake3 to {}-init in 351822d, matching CODING_STYLE.md and the existing brace-init lambda house style (network_callback.cpp:40, profiler/amd.cpp:362). Note: clang-format-15 reflows braced-init lambda bodies (it shows the identical delta on both of those existing sites), so I kept the documented hand-layout.

sources/benchmark/amd/blake3.cpp
yuzi-co: runKernel likewise switched to {}-init in 351822d.

sources/algo/blake3/opencl/blake3.cl
yuzi-co: Not intentional omission — every device helper carries `inline` on its own line as of 4286ddc: compress8, initVector, powHash, isLowerOrEqual. The two `__kernel` entry points (test_hash, search) deliberately omit it, since kernels can't be inline. If a specific function still looks unmarked, point me at it.

sources/common/kernel_generator/opencl.hpp
yuzi-co: clProgram is now consumed by the KATs after the #162 KernelGenerator conversion — each pulls multiple kernels by name from the single built program: sources/algo/blake3/opencl/tests/blake3_kat.cpp:87 and sources/algo/crypto/opencl/tests/blake3_kat.cpp:85 (cl::Kernel(generator.clProgram, name)). Production code still only reads clKernel.

PR #164, opened 2026-06-09
feat(amd): build ethash light cache on GPU

yuzi-co:
Implements #104.

## Summary

The AMD/OpenCL resolvers now honor `--ethash_light_cache_cpu` (default `true`) instead of hardcoding CPU light-cache generation. With `--ethash_light_cache_cpu=false`, only the 64-byte hashed seed is uploaded and the light cache is generated on-device, mirroring the existing NVIDIA CUDA path (`ethashBuildLightCache`).

Covers the whole ethash family — ethash, etchash, and all progpow variants (kawpow, meowpow, firopow, evrprogpow, progpow-quai, progpow-z) — via the shared `ResolverAmdEthash` / `ResolverAmdProgPOW` base classes.

## Changes

- **New kernel** `sources/algo/ethash/opencl/ethash_light_cache.cl` — single work-item. The light-cache build is strictly sequential (an iterated keccak-512 chain followed by 3 RandMemoHash rounds with an `i-1` read-after-write), so it cannot be parallelised — same as the CUDA `<<<1,1>>>` path. Reuses the existing `keccak_f1600`.
- **`common::opencl::Buffer::setFlags()`** — lets the light-cache buffer switch to `CL_MEM_READ_WRITE` when the kernel writes it (it is `CL_MEM_READ_ONLY` on the CPU-upload path).
- **Resolvers** — `updateContext()` reads the config flag; `updateMemory()` branches: CPU mode writes the full host cache (unchanged), GPU mode writes the 64-byte seed then runs `buildLightCache()`.
- **Tests** — `findNonceWithLightCacheGPU` for the AMD ethash and kawpow suites, mirroring the NVIDIA tests; ethash-family AMD fixtures pin `ethashBuildLightCacheCPU` for determinism.

## Why (it is not a speed feature)

The light-cache build is sequential by design, so on a single GPU it is ~9–10x slower than a CPU core — CPU build stays the default. The value is: no ~50–100 MB PCIe upload (just 64 bytes), no host-CPU load, and each GPU builds its own cache concurrently — useful on many-GPU rigs with a weak/contended CPU. This is the AMD counterpart of the option NVIDIA already exposes.

## Verification (RX 9070 XT, gfx1201)

- **Unit:** kawpow `findNonceWithLightCacheGPU` finds the exact expected nonce — end-to-end bit-exactness, since the progpow hash depends on the entire DAG, which depends on the entire light cache.
- **Live pool:** kawpow on `rvn.2miners.com` with `--ethash_light_cache_cpu=false` → log `Built light cache on GPU` → **Share valid (1 valid / 0 reject)**.

## Note on ethash + RDNA4

On RDNA4, plain **ethash** currently fails to produce valid shares regardless of CPU vs GPU light cache, due to a pre-existing DAG-build coverage bug (`ethash_build_dag` under-dispatches work-groups on gfx1201). That is independent of this PR — the CPU light-cache path is equally affected — and is handled separately. With that fix applied, the full ethash + kawpow GPU-light-cache test matrix is green on hardware. kawpow/progpow are unaffected and verified live as above.

### DIFF
--- a/sources/common/opencl/buffer_wrapper.hpp
+++ b/sources/common/opencl/buffer_wrapper.hpp
@@ -33,6 +33,12 @@ namespace common
             }
 
             inline
+            void setFlags(cl_mem_flags const _flags)
+            {
+                flags = _flags;
+            }
+
+            inline
             void setSize(size_t const _size)
             {
                 size = _size;
--- a/sources/resolver/amd/ethash.cpp
+++ b/sources/resolver/amd/ethash.cpp
@@ -101,8 +101,9 @@ bool resolver::ResolverAmdEthash::updateMemory(stratum::StratumJobInfo const& jo
     parameters.resultCache.free();
 
     ////////////////////////////////////////////////////////////////////////////
-    parameters.lightCache.setFlags(true == buildLightCacheOnCPU ? CL_MEM_READ_ONLY | CL_MEM_HOST_WRITE_ONLY
-                                                                : CL_MEM_READ_WRITE | CL_MEM_HOST_WRITE_ONLY);
+    parameters.lightCache.setFlags(
+        true == buildLightCacheOnCPU ? CL_MEM_READ_ONLY | CL_MEM_HOST_WRITE_ONLY
+                                     : CL_MEM_READ_WRITE | CL_MEM_HOST_WRITE_ONLY);
     parameters.lightCache.setSize(context.lightCache.size);
     parameters.dagCache.setSize(context.dagCache.size);
 
@@ -118,8 +119,8 @@ bool resolver::ResolverAmdEthash::updateMemory(stratum::StratumJobInfo const& jo
     if (true == buildLightCacheOnCPU)
     {
         if (false
-            == parameters.lightCache.write(
-                context.lightCache.hash, context.lightCache.size, clQueue[currentIndexStream]))
+            == parameters.lightCache
+                   .write(context.lightCache.hash, context.lightCache.size, clQueue[currentIndexStream]))
         {
             return false;
         }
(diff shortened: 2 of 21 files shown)


sources/common/opencl/buffer_wrapper.hpp
```diff
@@ -32,6 +32,12 @@ namespace common
                 size = _size;
             }
 
+            inline
+            void setFlags(cl_mem_flags const _flags)
```
luminousmining, line 36: never used ?

yuzi-co: It'\''s used in the GPU-build branch of `updateMemory()` — `sources/resolver/amd/ethash.cpp:104` and `sources/resolver/amd/progpow.cpp:96`:

```cpp
parameters.lightCache.setFlags(
    true == buildLightCacheOnCPU ? CL_MEM_READ_ONLY  | CL_MEM_HOST_WRITE_ONLY
                                 : CL_MEM_READ_WRITE | CL_MEM_HOST_WRITE_ONLY);
parameters.lightCache.setSize(context.lightCache.size);
```

`lightCache` is a persistent member, declared with the CPU-upload default `CL_MEM_READ_ONLY | CL_MEM_HOST_WRITE_ONLY`. When `--ethash_light_cache_cpu=false` the kernel *writes* the cache, so it has to become `CL_MEM_READ_WRITE` before `alloc()`. The flag is only known at runtime (it depends on the config) and the buffer isn'\''t reconstructed per job, so I set it the same way `setSize()` is set on the very next line — both mutate the member before the re-`alloc()`. Dropping `setFlags()` would leave the GPU path allocating a read-only cache the build kernel can'\''t write.

sources/resolver/amd/ethash.cpp
```diff
@@ -107,10 +116,26 @@ bool resolver::ResolverAmdEthash::updateMemory(stratum::StratumJobInfo const& jo
     }
 
     ////////////////////////////////////////////////////////////////////////////
-    if (false
-        == parameters.lightCache.write(context.lightCache.hash, context.lightCache.size, clQueue[currentIndexStream]))
+    if (true == buildLightCacheOnCPU)
```
luminousmining, line 119: Is the cleanup `algo::ethash::ContextGenerator::instance().free(algorithm);` missing?
Same logic applies to the other algorithms using `light cache on CPU`.

yuzi-co: The free isn't missing — updateMemory calls algo::ethash::ContextGenerator::instance().free(algorithm); at its tail, after buildDAG/buildSearch: ethash.cpp:148 (and the same for the others, e.g. progpow.cpp:140). It's unchanged by this PR, so it sits just below the diff hunk shown here — probably why it reads as absent. It runs for both the CPU-build and the new GPU-build path: the GPU branch only adds the on-device build between alloc() and that final free(), so the host context is released the same way it always was.

PR #172, opened 2026-06-11
feat(cpu): Blake3/Alephium CPU mining — first CPU algorithm

yuzi-co:
## Summary

Adds the first **CPU mining algorithm** by wiring the previously-unused CPU layer to
Blake3/Alephium support. A CPU is enumerated as a mining device alongside GPUs and runs
on the same global Blake3 pool/stratum path. The per-batch nonce scan is parallelized by a
**self-owned, persistent, pinned `std::thread` pool** — not OpenMP — exposing two new
controls, `--cpu_threads` and `--cpu_affinity`.

## Why a custom thread pool instead of OpenMP?

The scan originally used `#pragma omp parallel for`. That was replaced because:

1. **OpenMP did nothing on Windows.** The clang-cl cross build links no OpenMP runtime, so
   the pragma silently degraded to a **serial, single-core** scan — measured at **~1.5 MH/s
   on Windows vs ~12 MH/s on Linux** (libomp linked) on the same host, an ~8× gap that was
   purely serial-vs-multicore.
2. **No user control.** OpenMP chose the thread count itself; there was no way to set the
   worker count or pin workers to cores. The pool adds `--cpu_threads` / `--cpu_affinity`.
3. **Cross-platform runtime baggage.** OpenMP required libomp special-casing everywhere — a
   Linux `libomp-dev` apt-install + `libomp.so.5` runtime staging, a macOS Homebrew-libomp
   CMake probe, and a "best-effort serial" caveat on Windows.

The `std::jthread` pool fixes all three: real multicore on **every** platform, tunable, and
**zero** shipped runtime dependency. All OpenMP/libomp plumbing is deleted.

## What's included

- **Device layer**: `DEVICE_TYPE::CPU`, `DeviceCpu` (a no-SDK device modeled on
  `DeviceMocker`), enumerated by `DeviceManager` and dispatched through `setAlgorithm()`.
- **Resolver**: `ResolverCpuBlake3` — reuses the host `blake3::hashRef()` (double-BLAKE3
  Alephium PoW) for the nonce scan, fanned across the pool, with the same winner predicate /
  share-submit path as the GPU resolvers.
- **`CpuThreadPool`**: persistent worker threads pinned at startup; `parallelFor()` splits a
  batch into contiguous per-worker chunks and blocks until done. Pure helpers
  (`cpu_params.hpp`) and the pin wrapper (`cpu_affinity`) are unit-tested.
- **CLI**: `--cpu_threads` (pool size, default = all logical cores) and `--cpu_affinity`
  (hex core bitmask, default none). See [`documentation/CPU_TUNING.md`](documentation/CPU_TUNING.md).
- **Build**: CPU becomes an **orthogonal build axis** (`GPU=amd|nvidia|both|none` ×
  `CPU=ON|OFF`) across the Linux + Windows-cross Dockerfiles, `docker-build.ps1`, and CI;
  all `find_package(OpenMP)` / libomp install / staging removed.
- **Dashboard / defaults**: CPU shows as `CPU` (not `UNKNOWN`); CPU-appropriate occupancy so
  the hashrate displays out of the box. `--threads`/`--blocks` still override.

## Benchmarks (BLAKE3, live on HeroMiners, one config at a time)

Host: 16-core CPU + AMD RX 9070 XT.

| Config | Hashrate |
|---|---|
| Windows CPU only | **~11 MH/s**  *(was ~1.5 MH/s serial under OpenMP)* |
| Linux CPU only | ~10.7 MH/s |
| Windows GPU only | ~1.70 GH/s |
| Windows GPU + CPU (default) | ~0.4 GH/s ⚠️ |

**Headline:** Windows CPU now matches Linux (~11 vs ~10.7 MH/s) — the OpenMP Windows gap is
closed.

**GPU + CPU caveat:** running all-core CPU mining next to the GPU **starves the GPU's host
feeder thread**, collapsing it to ~0.4 GH/s. Reserve cores for the GPU with affinity —
`--cpu_threads=8 --cpu_affinity=0xFF` restores the full **1.70 GH/s GPU + ~7 MH/s CPU**.
Details in `CPU_TUNING.md`.

## Testing

- CI `unit-test` job green; new `CpuParams` (4), `CpuThreadPool` (2), and existing
  `ResolverBlake3Cpu` (3) suites pass.
- Live-verified on Alephium (HeroMiners) on Windows (CPU, GPU, and GPU+CPU) and Linux CPU
  (Docker) — jobs received, hashed, shares accepted.

## Notes

- **All** OpenMP/libomp plumbing is removed — no build, CPU or GPU, links it. CPU
  parallelism is now the in-process `std::thread` pool, which ships zero runtime deps.
- GPU-only builds (`BUILD_CPU=OFF`) simply don't compile the CPU resolver, so they're
  otherwise unchanged.
- History squashed to 3 commits: CPU device + resolver + thread pool · build axes · docs.


sources/algo/blake3/CMakeLists.txt
```diff
@@ -5,6 +5,14 @@ if (BUILD_EXE_MINER)
     target_sources(${MINER_EXE} PUBLIC
         ${HEADERS}
     )
+    if (BUILD_CPU)
+        # The CPU Blake3 resolver calls algo::blake3::hashRef (blake3_pow.cpp), which
```
luminousmining: remove comment

sources/device/cpu.cpp
```diff
@@ -0,0 +1,32 @@
...
+}
+
+
+uint32_t device::DeviceCpu::getMinimumKernelExecuted() const
+{
+    // A CPU batch is orders of magnitude slower than a GPU kernel, so the global
+    // default (100) is never reached between a pool's frequent job updates and the
+    // hashrate would read 0. Cap it low; honour a smaller user value if given.
+    constexpr uint32_t CPU_MINIMUM_KERNEL_EXECUTED{ 8u };
+    return std::min(device::Device::getMinimumKernelExecuted(), CPU_MINIMUM_KERNEL_EXECUTED);
```
luminousmining: use `common::min_limit` from custom.hpp

sources/device/cpu.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+void device::DeviceCpu::cleanUp()
+{
+    logInfo() << "Clean up device CPU";
+}
+
+
+uint32_t device::DeviceCpu::getMinimumKernelExecuted() const
+{
+    // A CPU batch is orders of magnitude slower than a GPU kernel, so the global
```
luminousmining: remove comment

sources/device/device.cpp
```diff
@@ -27,6 +27,9 @@
 #include <resolver/nvidia/meowpow.hpp>
 #include <resolver/nvidia/progpow.hpp>
 #include <resolver/nvidia/progpow_quai.hpp>
+#if defined(CPU_ENABLE)
+#include <resolver/cpu/blake3.hpp>
```
luminousmining: You can include it without any issue — the encapsulation / deactivation logic should be handled within the file itself `blake3.hpp` and not managed at every include site.

sources/device/device_manager.cpp
```diff
@@ -32,6 +32,9 @@
 #include <device/device_manager.hpp>
 #include <device/mocker.hpp>
 #include <device/nvidia.hpp>
+#if defined(CPU_ENABLE)
```
luminousmining: You can include it without any issue — the encapsulation / deactivation logic should be handled within the file itself `blake3.hpp` and not managed at every include site.

sources/device/device_manager.cpp
```diff
@@ -157,6 +172,13 @@ bool device::DeviceManager::initialize()
                     {
                         break;
                     }
+#endif
+#if defined(CPU_ENABLE)
+                    case device::DEVICE_TYPE::CPU:
+                    {
+                        // CPU mines the global --algo on the global pool (config.mining).
```
luminousmining: remove comment

sources/device/device_manager.cpp
```diff
@@ -299,6 +321,30 @@ bool device::DeviceManager::initializeMocker()
...
+    device::DeviceCpu* device{ NEW(device::DeviceCpu) };
+    if (nullptr == device) [[unlikely]]
+    {
+        logErr() << "Cannot alloc memory to create new CPU Device";
+        return false;
+    }
+
+    ///////////////////////////////////////////////////////////////////////////
+    device->deviceType = device::DEVICE_TYPE::CPU;
+    device->memoryAvailable = 0ull; // Blake3 has no DAG: the resolver never reads this.
```
luminousmining: remove comment

luminousmining (commented):
I have only looked at the first files.
I will do a more thorough analysis later.

sources/algo/blake3/CMakeLists.txt
yuzi-co: Removed.

sources/device/cpu.cpp
yuzi-co: Removed.

sources/device/device_manager.cpp
yuzi-co: Removed.

sources/device/device_manager.cpp
yuzi-co: Removed.

sources/device/device.cpp
yuzi-co: Done — `resolver/cpu/blake3.hpp` now wraps its body in `#if defined(CPU_ENABLE)` (same pattern as `device/cpu.hpp`), so this include is unconditional and the header compiles to nothing when CPU is disabled.

sources/device/device_manager.cpp
yuzi-co: `device/cpu.hpp` already self-guards with `CPU_ENABLE`; removed the guard at the include site so the include is unconditional, with deactivation handled inside the header.

sources/device/cpu.cpp
yuzi-co: Used `common::max_limit` rather than `min_limit`. In custom.hpp `max_limit(v,m)=v<=m?v:m` is the ceiling (== std::min) and `min_limit(v,m)=v>=m?v:m` is the floor (== std::max). This needs to cap the threshold at 8 — `std::min(default, 8)` — so a slow CPU still publishes stats; `min_limit(100,8)` would return 100 and defeat the cap. Same direction as the existing `max_limit(data.count, MAX_RESULT)` clamps. Happy to switch if you intended different semantics.

sources/device/device.cpp
```diff
@@ -507,12 +516,17 @@ void device::Device::increaseShare(bool const isValid)
 }
 
 
+uint32_t device::Device::getMinimumKernelExecuted() const
+{
+    return common::Config::instance().occupancy.kernelMinimunExecuteNeeded.value_or(100u);
```
luminousmining:
```cpp
common::Config& config{ common::Config::instance() };
uint32_t const  executeCount{ .occupancy.kernelMinimunExecuteNeeded };
return executeCount;
```

Does not need `.value_or(100u)` the default value is `100`

sources/device/device_manager.cpp
```diff
@@ -299,6 +318,30 @@ bool device::DeviceManager::initializeMocker()
 }
 #endif
 
+#if defined(CPU_ENABLE)
```
luminousmining, line 322: Need 2 empty lines

sources/device/device_manager.cpp
```diff
@@ -299,6 +318,30 @@ bool device::DeviceManager::initializeMocker()
...
+
+    ///////////////////////////////////////////////////////////////////////////
+    logInfo() << "CPU[" << devices.size() << "] enabled";
+    devices.push_back(device);
+
+    return true;
+}
+#endif
+
 #if defined(CUDA_ENABLE)
```
luminousmining, line 347: Need 2 empty lines

sources/resolver/cpu/cpu_params.hpp
```diff
@@ -0,0 +1,100 @@
...
+#include <algorithm>
+#include <bit>
+#include <cstdint>
+#include <limits>
+#include <optional>
+#include <string>
+#include <utility>
+
+
+namespace resolver::cpu_detail
```
luminousmining: wrong define namespace look resolver `nvidia` or `amd`
namespace `resolver` { namespace `cpu`

sources/resolver/cpu/cpu_params.hpp
luminousmining, line 1: Wrong function signature:

```cpp
inline
type function_name(parameters)
```

sources/resolver/cpu/tests/cpu_params.cpp
```diff
@@ -0,0 +1,61 @@
+#include <gtest/gtest.h>
+
+#include <resolver/cpu/cpu_params.hpp>
+
+
+using namespace resolver::cpu_detail;
```
luminousmining: never declare `using` namespace out of scope.
Declare in function or use resolver::X::

sources/resolver/cpu/blake3.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include <resolver/cpu/blake3.hpp>
+#include <resolver/cpu/cpu_params.hpp>
+
+
+// Default CPU occupancy: DEFAULT_THREADS * DEFAULT_BLOCKS = 262144 nonces scanned per
+// executeSync(). Big enough to amortize the thread-pool dispatch per batch (tiny batches
+// measurably cut multi-core throughput), small enough that ~8 batches still complete
+// between a pool's frequent jobs so the hashrate displays (see
+// DeviceCpu::getMinimumKernelExecuted). Overridden by --threads / --blocks.
+namespace
```
luminousmining: no namespace empty

sources/resolver/cpu/cpu.cpp
```diff
@@ -4,4 +4,15 @@
...
+    ////////////////////////////////////////////////////////////////////////////
+    common::Config const& config{ common::Config::instance() };
+
+    // blocks * threads is the nonce count scanned per executeSync() call. Unlike the
+    // GPU paths there is no warp/group-size constraint on the CPU, so honour any
+    // user --threads/--blocks verbatim and otherwise fall back to the caller's
+    // defaults (kept small so each batch is short: the loop reacts quickly to new
+    // jobs and the hashrate counter accumulates enough batches between job resets to
+    // display).
+    setThreads(std::nullopt != config.occupancy.threads ? *config.occupancy.threads : defaultThreads);
```
luminousmining: declare 2 variables for threads and blocks please

sources/resolver/cpu/cpu_affinity.cpp
```diff
@@ -0,0 +1,35 @@
...
+#elif defined(__linux__)
+#include <pthread.h>
+#include <sched.h>
+#endif
+
+
+bool resolver::pinThisThreadToCore([[maybe_unused]] uint32_t const coreIndex)
+{
+#if defined(_WIN32)
+    DWORD_PTR const mask{ static_cast<DWORD_PTR>(1ull << coreIndex) };
```
luminousmining: Can you create a custom cast in `cast.hpp` ?

sources/resolver/cpu/cpu_affinity.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <cstdint>
+
+
+namespace resolver
+{
+    // Pin the CALLING thread to a single logical core. Returns true on success.
+    // macOS (and any platform without a hard-affinity API): best-effort no-op, returns false.
+    bool pinThisThreadToCore(uint32_t coreIndex);
```
luminousmining: uint32_t const coreIndex

sources/resolver/cpu/cpu_params.hpp
luminousmining, line 1: use `cast.hpp` instead of `static_cast`

sources/resolver/cpu/thread_pool.cpp
```diff
@@ -0,0 +1,117 @@
...
+                });
+        }
+    }
+}
+
+
+resolver::CpuThreadPool::~CpuThreadPool()
+{
+    {
+        std::lock_guard<std::mutex> const lock{ mutex };
```
luminousmining: use custom UNIQUE_LOCK from `custom.hpp`

sources/resolver/cpu/thread_pool.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <vector>
+
+
+namespace resolver
+{
+    // A fixed-size pool of persistent worker threads, each optionally pinned to a logical
+    // core at startup. parallelFor() splits [0, count) into contiguous per-worker chunks and
+    // blocks until all workers finish. Mining hits are astronomically rare, so callers guard
+    // any shared write with their own mutex inside the chunk function.
+    class CpuThreadPool
```
luminousmining: I understand the use of `jthread` but in this project I prefer to stay close to the `boost` framework, and to keep consistency with `device_manager` can you study the possibility of switching to `boost::thread` or `boost::scoped_thread` or other alternatives — I will leave the research to you.
Taking into account my remark that 1 thread could be a `devicecpu`, `CpuThreadPool` would therefore no longer be needed.

sources/common/cast.hpp
```diff
@@ -71,3 +71,11 @@
 ////////////////////////////////////////////////////////////////////////////////
 #define castPtrHash512(value) reinterpret_cast<algo::hash512*>(value)
 #endif // !__LIB_CUDA
+
+
+#if defined(_WIN32)
+////////////////////////////////////////////////////////////////////////////////
```
luminousmining: remove comment

sources/device/device.cpp
```diff
@@ -507,12 +516,17 @@ void device::Device::increaseShare(bool const isValid)
 }
 
 
+uint32_t device::Device::getMinimumKernelExecuted() const
+{
+    return common::Config::instance().occupancy.kernelMinimunExecuteNeeded;
```
luminousmining: add tmp var

sources/device/device_manager.cpp
```diff
@@ -32,6 +32,7 @@
 #include <device/device_manager.hpp>
 #include <device/mocker.hpp>
 #include <device/nvidia.hpp>
+#include <device/cpu.hpp>
```
luminousmining: alpha order

sources/resolver/cpu/cpu_params.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+        // Effective worker count: explicit --cpu_threads (>0) wins; else popcount(mask) when an
+        // affinity mask is given; else the machine concurrency; never below 1.
+        inline uint32_t
+        resolveWorkerCount(
+            std::optional<uint32_t> const threads,
+            uint64_t const                mask,
+            uint32_t const                hardwareConcurrency)
+        {
+            if (threads.has_value() && 0u < *threads)
```
luminousmining: prefere compare `std::nullopt`

sources/resolver/cpu/cpu_params.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include <common/cast.hpp>
+
+
+namespace resolver
+{
+    namespace cpu
+    {
+        // 0-based bit position of the k-th set bit of mask, or 64 if fewer than k+1 bits set.
+        inline
+        uint32_t nthSetBit(uint64_t const mask, uint32_t k)
```
luminousmining: This is not a function specific to the CPU but rather bit manipulation.
Place it in `bitwise.hpp`.

sources/resolver/cpu/cpu_params.hpp
```diff
@@ -0,0 +1,110 @@
...
+            uint64_t const base{ total / workerCount };
+            uint64_t const rem{ total % workerCount };
+            uint64_t const lo{ index * base + std::min<uint64_t>(index, rem) };
+            uint64_t const extra{ (index < rem) ? 1ull : 0ull };
+            return { lo, lo + base + extra };
+        }
+
+        // Parse a hex affinity mask ("0xFF" or "FF") to uint64. Empty or invalid input -> 0.
+        inline
+        uint64_t parseHexMask(std::string const& text)
```
luminousmining: This function should be moved to a `common` file, probably `bitwise.hpp` — it is not strictly tied to the CPU.
It can be renamed `hexToDecimal`, and we could even provide a `template` to be compatible with 8/16/32/64 bits.

sources/resolver/cpu/cpu_affinity.cpp
```diff
@@ -0,0 +1,36 @@
...
+#if defined(_WIN32)
+    DWORD_PTR const mask{ castDWORDPTR(1ull << coreIndex) };
+    return 0 != SetThreadAffinityMask(GetCurrentThread(), mask);
+#elif defined(__linux__)
+    cpu_set_t set;
+    CPU_ZERO(&set);
+    CPU_SET(coreIndex, &set);
+    return 0 == pthread_setaffinity_np(pthread_self(), sizeof(set), &set);
+#else
+    static bool warned{ false };
```
luminousmining: We are not going to implement a `static` variable for this warning — I prefer to handle it on every pass.

sources/resolver/cpu/cpu_affinity.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include <pthread.h>
+#include <sched.h>
+#endif
+
+
+bool resolver::pinThisThreadToCore([[maybe_unused]] uint32_t const coreIndex)
+{
+#if defined(_WIN32)
+    DWORD_PTR const mask{ castDWORDPTR(1ull << coreIndex) };
+    return 0 != SetThreadAffinityMask(GetCurrentThread(), mask);
```
luminousmining: I prefere tmp var `bool const success`

sources/resolver/cpu/cpu_affinity.cpp
```diff
@@ -0,0 +1,36 @@
...
+#elif defined(__linux__)
+    cpu_set_t set;
+    CPU_ZERO(&set);
+    CPU_SET(coreIndex, &set);
+    return 0 == pthread_setaffinity_np(pthread_self(), sizeof(set), &set);
```
luminousmining: I prefere tmp var `bool const success`

sources/resolver/cpu/cpu_affinity.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <cstdint>
+
+
+namespace resolver
```
luminousmining, line 6: missing `namespace cpu`

sources/resolver/cpu/thread_pool.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include <atomic>
+#include <cstdint>
+#include <functional>
+#include <vector>
+
+#include <boost/thread.hpp>
+
+
+namespace resolver
+{
```
luminousmining, line 13: missing `namespace cpu {}`

sources/resolver/cpu/tests/blake3.cpp
```diff
@@ -0,0 +1,134 @@
+#include <gtest/gtest.h>
+
+#include <algo/hash.hpp>
+#include <algo/hash_utils.hpp>
+#include <common/mocker/stratum.hpp>
+#include <resolver/cpu/blake3.hpp>
+#include <stratum/job_info.hpp>
+
+
+namespace
```
luminousmining: Namespace empty is forbidden

sources/resolver/cpu/tests/cpu_params.cpp
```diff
@@ -0,0 +1,66 @@
...
+    EXPECT_EQ(1u, resolveWorkerCount(std::nullopt, 0ull, 0u));                  // floor at 1
+    EXPECT_EQ(8u, resolveWorkerCount(std::optional<uint32_t>{ 0u }, 0ull, 8u)); // 0 == unset
+}
+
+
+TEST(CpuParams, chunkRangeCoversExactly)
+{
+    using namespace resolver::cpu;
+
+    for (uint64_t const total : { 0ull, 1ull, 7ull, 100ull, 262144ull })
```
luminousmining: Create tmp var contains `{ 0ull, 1ull, 7ull, 100ull, 262144ull }`

sources/resolver/cpu/tests/cpu_params.cpp
```diff
@@ -0,0 +1,66 @@
...
+    {
+        for (uint32_t const n : { 1u, 3u, 8u })
```
luminousmining: Create var to contain `{ 1u, 3u, 8u }`.

luminousmining: `n` ? explicit name please

sources/resolver/cpu/blake3.cpp
```diff
@@ -0,0 +1,274 @@
...
+#include <resolver/cpu/cpu_params.hpp>
+
+
+resolver::ResolverCpuBlake3::PoolConfig resolver::ResolverCpuBlake3::resolvePoolConfig()
+{
+    common::Config const& config{ common::Config::instance() };
+
+    // Parse the affinity mask exactly once: the worker-count resolution needs it (popcount
+    // when --cpu_threads is unset) and the pool needs it for pinning.
+    uint64_t const mask{ config.cpu.affinity.has_value() ? resolver::cpu::parseHexMask(*config.cpu.affinity) : 0ull };
```
luminousmining: prefere `std::null opt`

sources/resolver/cpu/thread_pool.hpp
```diff
@@ -0,0 +1,62 @@
...
+    // rest of the threading in the project (see device_manager).
+    //
+    // run() dispatches and blocks until the batch finishes (used by executeSync, tests, debug).
+    // runAsync() dispatches and returns immediately; the caller drains it later with wait().
+    // This is the CPU mirror of the GPU two-stream double-buffer: the device reads/submits the
+    // previous batch and syncs with stratum while the workers compute the next one.
+    class CpuThreadPool
+    {
+      public:
+        using ChunkFn = std::function<void(uint64_t lo, uint64_t hi, uint32_t workerIndex)>;
```
luminousmining: look `stratum.hpp` alias for function must start with `callback`
Exemples :
```cpp
        using callbackUpdateJob = std::function<void(uint32_t const stratumUUID, StratumJobInfo const&)>;
        using callbackShareStatus =
            std::function<void(bool const isValid, uint32_t const requestID, uint32_t const stratumUUID)>;
```

sources/resolver/cpu/thread_pool.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+        CpuThreadPool(uint32_t workerCount, uint64_t affinityMask);
+        ~CpuThreadPool();
+
+        CpuThreadPool(CpuThreadPool const&) = delete;
+        CpuThreadPool& operator=(CpuThreadPool const&) = delete;
+
+        void setCallback(ChunkFn const& fn);
+        // grain == 0 selects the default ceil(count / workers) (one slice per worker).
```
luminousmining: can remove comment

sources/resolver/cpu/thread_pool.hpp
```diff
@@ -0,0 +1,62 @@
...
+        void runAsync(uint64_t count, uint64_t grain = 0ull);
+        void wait();
+        void run(uint64_t count, uint64_t grain = 0ull);
+
+      private:
+        void workerLoop(uint32_t index);
+
+        uint32_t                   poolSize;
```
luminousmining: Global message, variable should be all time initialize by default.

sources/resolver/cpu/thread_pool.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+#include <common/cast.hpp>
+#include <common/custom.hpp>
+#include <resolver/cpu/cpu_affinity.hpp>
+#include <resolver/cpu/cpu_params.hpp>
+#include <resolver/cpu/thread_pool.hpp>
+
+
+resolver::CpuThreadPool::CpuThreadPool(uint32_t const workerCount, uint64_t const affinityMask)
+    : poolSize{ (0u < workerCount) ? workerCount : 1u }, mask{ affinityMask }
```
luminousmining, line 14: : poolSize{ (0u < workerCount) ? workerCount : 1u },
  mask{ affinityMask }

If clang-format accept

sources/resolver/cpu/thread_pool.cpp
```diff
@@ -0,0 +1,173 @@
...
+{
+    {
+        UNIQUE_LOCK(mutex);
+        stopRequested = true;
+    }
+    cvWork.notify_all();
+    // boost::thread does not auto-join on destruction: join each worker explicitly.
+    for (boost::thread& worker : workers)
+    {
+        if (true == worker.joinable())
```
luminousmining, line 46: you can use `.interrupt()`

sources/resolver/cpu/thread_pool.cpp
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+
+void resolver::CpuThreadPool::workerLoop(uint32_t const index)
+{
+    if (0ull != mask)
+    {
+        uint32_t const population{ castU32(std::popcount(mask)) };
+        resolver::pinThisThreadToCore(resolver::cpu::nthSetBit(mask, index % population));
```
luminousmining: tmp var for `resolver::cpu::nthSetBit(mask, index % population)`

sources/resolver/cpu/thread_pool.cpp
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    uint64_t lastGeneration{ 0ull };
+    for (;;)
```
luminousmining: no `infinity` loop.
Define a explicit condition to stop this loop.

sources/resolver/cpu/thread_pool.cpp
```diff
@@ -0,0 +1,173 @@
...
+            jobGrain = sliceGrain;
+        }
+
+        // Pull grain-sized slices from the shared cursor until [0, jobTotal) is drained. The
+        // fetch_add is an atomic read-modify-write, so it always sees the latest value in the
+        // modification order and no slice is ever handed out twice -- relaxed ordering is
+        // enough here because the slices are independent (each worker only touches its own
+        // [lo, hi) nonces; the lone shared write is the hit append, which the callback guards
+        // with its own mutex).
+        for (;;)
```
luminousmining: no `infinity` loop.
Define a explicit condition to stop this loop.

sources/resolver/cpu/thread_pool.cpp
```diff
@@ -0,0 +1,173 @@
...
+    uint64_t lastGeneration{ 0ull };
+    for (;;)
+    {
+        uint64_t jobTotal{ 0ull };
+        uint64_t jobGrain{ 1ull };
+        {
+            UNIQUE_LOCK(mutex);
+            cvWork.wait(
+                lock,
+                [&]
```
luminousmining: define this lambda out this call

sources/resolver/cpu/thread_pool.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+    uint64_t lastGeneration{ 0ull };
+    for (;;)
+    {
+        uint64_t jobTotal{ 0ull };
+        uint64_t jobGrain{ 1ull };
+        {
+            UNIQUE_LOCK(mutex);
+            cvWork.wait(
+                lock,
```
luminousmining: `lock `? Where is it define ?

sources/resolver/cpu/blake3.cpp
```diff
@@ -0,0 +1,274 @@
...
+    uint32_t const workers{
+        resolver::cpu::resolveWorkerCount(config.cpu.threads, mask, std::thread::hardware_concurrency())
+    };
+    return PoolConfig{ workers, mask };
+}
+
+
+resolver::ResolverCpuBlake3::ResolverCpuBlake3(PoolConfig const poolConfig)
+    : resolver::ResolverCpu(), threadPool{ poolConfig.workerCount, poolConfig.affinityMask }
```
luminousmining, line 45:
```cpp
: resolver::ResolverCpu(),
  threadPool{ poolConfig.workerCount, poolConfig.affinityMask }
```

sources/resolver/cpu/blake3.cpp
```diff
@@ -0,0 +1,274 @@
...
+    // launch the next batch into the now-idle one and return immediately. While the device
+    // reads/submits the harvested buffer and syncs with stratum, the workers keep computing.
+    if (true == inFlight)
+    {
+        threadPool.wait();
+        harvest(batch[currentIndex], jobInfo);
+    }
+
+    ////////////////////////////////////////////////////////////////////////////
+    currentIndex ^= 1u;
```
luminousmining: `void     swapIndexStream();` use this function.
You can use `currentIndexStream` directly

sources/resolver/cpu/blake3.cpp
```diff
@@ -0,0 +1,274 @@
...
+    Batch& next{ batch[currentIndex] };
+    prepareBatch(next, jobInfo);
+
+    uint32_t const launchIndex{ currentIndex };
+    threadPool.setCallback(
+        [this, launchIndex](uint64_t const lo, uint64_t const hi, uint32_t const /*workerIndex*/)
```
luminousmining: `uint32_t const /*workerIndex*/` should be delete or `[[maybe_unused]]`

sources/resolver/cpu/blake3.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    ////////////////////////////////////////////////////////////////////////////
+    uint64_t const count{ castU64(getBlocks()) * castU64(getThreads()) };
+    Batch&         current{ batch[currentIndex] };
+
+    prepareBatch(current, jobInfo);
+
+    // Fan the nonce batch across the pinned worker pool and block until it finishes: the
+    // synchronous path is the one tests and debugging rely on.
+    threadPool.setCallback(
```
luminousmining: We should not have to reassign the `callback` on every `execute` call.

sources/resolver/cpu/blake3.cpp
```diff
@@ -0,0 +1,274 @@
...
+        harvest(batch[currentIndex], jobInfo);
+    }
+
+    ////////////////////////////////////////////////////////////////////////////
+    currentIndex ^= 1u;
+    Batch& next{ batch[currentIndex] };
+    prepareBatch(next, jobInfo);
+
+    uint32_t const launchIndex{ currentIndex };
+    threadPool.setCallback(
```
luminousmining: We should not have to reassign the `callback` on every `execute` call.

sources/resolver/cpu/blake3.cpp
```diff
@@ -0,0 +1,274 @@
...
+        {
+            for (uint32_t i{ 0u }; i < resultShare.count; ++i)
+            {
+                // Zero-pad to 16 hex (8 bytes): the pool recomputes the 24-byte nonce, so leading zeros matter.
+                std::stringstream nonceHexa;
+                nonceHexa << std::setw(16) << std::setfill('0') << std::hex << resultShare.nonces[i];
+
+                std::string nonceStr{ nonceHexa.str() };
+
+                while (nonceStr.size() < 48)
```
luminousmining: magic number, should be value from `hash.hpp` ? Exemple `also::LEN_HASH_512_WORD_8` ?

luminousmining: Doesn't `std::setfill('0')` already fill with zeros?

sources/resolver/cpu/thread_pool.cpp
```diff
@@ -0,0 +1,173 @@
...
+    // Inline path (poolSize <= 1) already finished synchronously inside runAsync().
+    if (1u >= poolSize)
+    {
+        return;
+    }
+
+    UNIQUE_LOCK(mutex);
+    cvDone.wait(
+        lock,
+        [&]
```
luminousmining: create lambda `auto isRemaining`

sources/resolver/cpu/thread_pool.hpp
```diff
@@ -0,0 +1,62 @@
...
+        uint64_t                   mask;
+        ChunkFn                    job{};
```
luminousmining: `cbJob` cf `ChunkFn` rename

sources/resolver/cpu/thread_pool.hpp
```diff
@@ -0,0 +1,62 @@
...
+        uint64_t                   total{ 0ull };
+        uint64_t                   sliceGrain{ 1ull };
+        std::atomic<uint64_t>      cursor{ 0ull };
```
luminousmining: You should use `AtomicCounter` sources/common/atomic_counter.hpp
You can add new function `setMemoryOrder` to define if the boost::memory_order.
`current.fetch_add(value, boost::memory_order::seq_cst);` -> `current.fetch_add(value, memoryOrder);`

sources/resolver/cpu/thread_pool.cpp
```diff
@@ -0,0 +1,173 @@
...
+        // [lo, hi) nonces; the lone shared write is the hit append, which the callback guards
+        // with its own mutex).
+        for (;;)
+        {
+            uint64_t const lo{ cursor.fetch_add(jobGrain, std::memory_order_relaxed) };
+            if (lo >= jobTotal)
+            {
+                break;
+            }
+            uint64_t const hi{ std::min<uint64_t>(lo + jobGrain, jobTotal) };
```
luminousmining: use `math.hpp` -> `MIN()`

sources/resolver/cpu/thread_pool.cpp
```diff
@@ -0,0 +1,173 @@
...
+    : poolSize{ (0u < workerCount) ? workerCount : 1u }, mask{ affinityMask }
+{
+    // poolSize == 1 runs inline in run(); no worker threads are spawned.
+    if (1u < poolSize)
+    {
+        workers.reserve(poolSize);
+        for (uint32_t i{ 0u }; i < poolSize; ++i)
+        {
+            workers.emplace_back(
+                [this, i]
```
luminousmining, line 27: I prefere lambda declared before, if lambda can be define in one line you can dot it ;)

sources/resolver/cpu/thread_pool.hpp
yuzi-co: Done in `3a7b8eb` — renamed the alias `ChunkFn` → `callbackJob` to match the `callback*` convention in `stratum.hpp`.

sources/resolver/cpu/thread_pool.hpp
yuzi-co: Done in `3a7b8eb` — removed.

sources/resolver/cpu/thread_pool.hpp
yuzi-co: Done in `3a7b8eb` — `poolSize{ 1u }` and `mask{ 0ull }` are now default-initialized.

sources/resolver/cpu/thread_pool.cpp
yuzi-co: Done in `3a7b8eb` — hoisted into a `core` local before `pinThisThreadToCore`.

sources/resolver/cpu/thread_pool.cpp
yuzi-co: Done in `3a7b8eb` — the outer loop is now `while (false == stopRequested.load(std::memory_order_acquire))` (made `stopRequested` a `std::atomic<bool>` for the lockless read).

sources/resolver/cpu/thread_pool.cpp
yuzi-co: Done in `3a7b8eb` — the cursor drain is now `for (uint64_t lo{ cursor.fetch_add(...) }; lo < jobTotal; lo = cursor.fetch_add(...))`, no more `for(;;)`/break.

sources/resolver/cpu/thread_pool.cpp
yuzi-co: Done in `3a7b8eb` — hoisted to a named `hasWork` lambda before the `wait()` call.

sources/resolver/cpu/thread_pool.cpp
yuzi-co: Done in `3a7b8eb` — hoisted to `auto const isRemaining{ [&]() { return 0u == remaining; } };`.

sources/resolver/cpu/thread_pool.hpp
yuzi-co: Done in `3a7b8eb` — member `job` → `cbJob`.

sources/resolver/cpu/blake3.cpp
yuzi-co: Done in `3a7b8eb` — dropped the private `currentIndex`; `executeSync`/`executeAsync` now use the base-class `currentIndexStream` + `swapIndexStream()`.

sources/resolver/cpu/blake3.cpp
yuzi-co: Done in `3a7b8eb` — now `[[maybe_unused]] uint32_t const workerIndex`.

sources/resolver/cpu/blake3.cpp
yuzi-co: Done in `3a7b8eb` — the callback is now installed once in the constructor (it scans into `batch[currentIndexStream]`); `executeSync`/`executeAsync` only flip the index, never reassign.

sources/resolver/cpu/blake3.cpp
yuzi-co: Done in `3a7b8eb` — see above: the callback is set once in the constructor now; no per-execute reassignment.

sources/resolver/cpu/thread_pool.cpp
yuzi-co: clang-format 15.0.7 packs the initializer list back onto one line (there is no `PackConstructorInitializers` override, so it defaults to BinPack), so the two-line form fails `format-check`. I left it as the single line the formatter produces — happy to add `PackConstructorInitializers: NextLine` (or similar) to `.clang-format` instead if you want one-initializer-per-line project-wide.

sources/resolver/cpu/blake3.cpp
yuzi-co: Same as the thread_pool constructor: clang-format 15.0.7 rejoins this onto one line, so I kept the formatter output.

sources/resolver/cpu/thread_pool.cpp
yuzi-co: I kept the explicit `stopRequested` flag rather than `.interrupt()`. `boost::thread::interrupt()` works by throwing `boost::thread_interrupted` at interruption points, which adds an exception path through the worker hot loop; the pool is also always quiesced before teardown (`~ResolverCpuBlake3` drains any in-flight batch first), so the flag + `notify_all` is enough and keeps shutdown exception-free. Glad to switch to `.interrupt()` if you prefer it for consistency.

sources/resolver/cpu/thread_pool.cpp
yuzi-co: `lock` is declared by the `UNIQUE_LOCK(mutex)` macro on the line above — `common/custom.hpp`: `#define UNIQUE_LOCK(mtxName) boost::unique_lock<boost::mutex> lock{ mtxName };`. Same macro used throughout the codebase.

sources/resolver/cpu/thread_pool.hpp
yuzi-co: `AtomicCounter` currently exposes only `void add()`, but the cursor needs the *previous* value returned by the RMW (`lo = cursor.fetch_add(grain)`), which `AtomicCounter` cannot provide today. To adopt it I would add a value-returning `fetchAdd()` plus a `setMemoryOrder()` to `common/atomic_counter.hpp` (the only other user is `device.hpp`). OK to extend the shared type that way?

sources/resolver/cpu/thread_pool.cpp
yuzi-co: I could not find `sources/common/math.hpp` or a `MIN()` anywhere on this branch (or on `main`). Did you mean a helper landing separately, or `common/custom.hpp`'s `max_limit`/`min_limit` (those clamp, not min)? Kept `std::min` for now — point me at the intended one and I will switch.

sources/resolver/cpu/blake3.cpp
yuzi-co: There is no matching constant in `hash.hpp`: the Alephium nonce is 24 bytes / 48 hex (and the 8-byte search value is 16 hex), neither of which maps to a `LEN_HASH_*` (e.g. `LEN_HASH_512_WORD_8` is 64 bytes). I named them locally instead — `SEARCH_NONCE_HEX_LENGTH` (16) and `FULL_NONCE_HEX_LENGTH` (48).

yuzi-co: Right — `setw(16)` + `setfill('0')` left-pads the 8-byte search value to 16 hex. The separate `while (size() < 48)` loop does a different job: right-extending those 16 hex to the full 24-byte (48 hex) nonce the pool recomputes. Kept both, now via the named constants.

sources/resolver/cpu/thread_pool.cpp
yuzi-co: clang-format 15.0.7 expands lambda bodies to Allman (braces on their own line), so the one-line form is not reachable. This lambda is only a trampoline to `workerLoop(i)` capturing the per-iteration `i`, so there is nothing to hoist out of the loop — I left it as the formatter multi-line form.

sources/resolver/cpu/thread_pool.cpp
yuzi-co: Done in `6389b21` — now uses `algo::min()` from `sources/algo/math.hpp` (I had been looking under `common/`). Thanks for the pointer.

sources/resolver/cpu/thread_pool.hpp
yuzi-co: Done in `6389b21` — `AtomicCounter::add` (and `sub`) now return the RMW result and there is a `setMemoryOrder()` (plus a `store()` for resetting). `get`/`sub`/`store` all honor the configurable `memoryOrder`, which defaults to `seq_cst` so the `device.cpp` synchronizer counters are unchanged. The cursor is now `common::AtomicCounter<uint64_t>` set to relaxed ordering.

sources/resolver/cpu/thread_pool.cpp
yuzi-co: Done in `6389b21` — switched to `worker.interrupt()` in the destructor; the worker loop now runs `while (false == boost::this_thread::interruption_requested())` and catches the `boost::thread_interrupted` thrown out of the interruptible `cvWork.wait()`. The stop flag is gone.

sources/algo/tests/bitwise.cpp
```diff
@@ -0,0 +1,42 @@
...
+    EXPECT_EQ(0xFFFFFFFFFFFFFFFFull, algo::hexToDecimal<uint64_t>("0xFFFFFFFFFFFFFFFF")); // 16 hex digits: fits
+    EXPECT_EQ(0ull, algo::hexToDecimal<uint64_t>("0x1FFFFFFFFFFFFFFFF"));                 // 17 sig digits: overflow
+}
+
+
+// The template must clamp to each width: a value needing more nibbles than T holds is rejected
+// (returns 0), while the widest value that fits is parsed exactly.
+TEST(Bitwise, hexToDecimalWidths)
+{
+    EXPECT_EQ(static_cast<uint8_t>(0xABu), algo::hexToDecimal<uint8_t>("AB"));
```
luminousmining: cast.hpp

sources/common/cast.hpp
```diff
@@ -71,3 +71,9 @@
...
```
luminousmining: remove this line `///////`

sources/device/device.cpp
```diff
@@ -507,12 +516,18 @@ void device::Device::increaseShare(bool const isValid)
 }
 
 
+uint32_t device::Device::getMinimumKernelExecuted() const
+{
+    uint32_t const executeCount{ common::Config::instance().occupancy.kernelMinimunExecuteNeeded };
```
luminousmining:
```cpp
common::Config const& config{ common::Config::instance() };
return config..occupancy.kernelMinimunExecuteNeeded;
```

sources/resolver/cpu/tests/thread_pool.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include <gtest/gtest.h>
+
+#include <resolver/cpu/thread_pool.hpp>
+
+
+// run() must visit every index in [0, count) exactly once, across worker counts and
+// batch sizes including count < workers and count not divisible by workers.
+TEST(CpuThreadPool, runCoversAllIndices)
+{
+    for (uint32_t const workers : { 1u, 2u, 4u, 8u })
```
luminousmining: create variable to assign `{ 1u, 2u, 4u, 8u }`

sources/resolver/cpu/tests/thread_pool.cpp
```diff
@@ -0,0 +1,131 @@
...
+    {
+        for (uint64_t const count : { 0ull, 1ull, 5ull, 1000ull })
```
luminousmining: create variable to assign `{ 0ull, 1ull, 5ull, 1000ull }`

sources/resolver/cpu/tests/thread_pool.cpp
```diff
@@ -0,0 +1,131 @@
...
+                    for (uint64_t i{ lo }; i < hi; ++i)
+                    {
+                        seen[static_cast<size_t>(i)].fetch_add(1);
+                    }
+                });
+            pool.run(count);
+
+            for (uint64_t i{ 0ull }; i < count; ++i)
+            {
+                EXPECT_EQ(1, seen[static_cast<size_t>(i)].load())
```
luminousmining: cast.hpp or count -> `size_t`

sources/resolver/cpu/tests/thread_pool.cpp
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+}
+
+
+// A small explicit grain turns the single cursor into work-stealing: every index in
+// [0, count) must still be visited exactly once regardless of how finely it is sliced.
+TEST(CpuThreadPool, smallGrainStillCoversAllIndicesExactlyOnce)
+{
+    for (uint32_t const workers : { 2u, 4u, 8u })
```
luminousmining: same define variable for this list

sources/resolver/cpu/tests/thread_pool.cpp
```diff
@@ -0,0 +1,131 @@
...
+    {
+        for (uint64_t const grain : { 1ull, 3ull, 7ull })
```
luminousmining: same define variable for this list

sources/resolver/cpu/tests/thread_pool.cpp
```diff
@@ -0,0 +1,131 @@
...
+                    for (uint64_t i{ lo }; i < hi; ++i)
+                    {
+                        seen[static_cast<size_t>(i)].fetch_add(1);
+                    }
+                });
+            pool.run(count, grain);
+
+            for (uint64_t i{ 0ull }; i < count; ++i)
+            {
+                EXPECT_EQ(1, seen[static_cast<size_t>(i)].load())
```
luminousmining: same cast.hpp or define count at size_t

sources/resolver/cpu/tests/thread_pool.cpp
```diff
@@ -0,0 +1,131 @@
...
+                seen[static_cast<size_t>(i)].fetch_add(1);
+            }
+        });
+
+    pool.runAsync(count);
+    pool.wait();
+
+    for (uint64_t i{ 0ull }; i < count; ++i)
+    {
+        EXPECT_EQ(1, seen[static_cast<size_t>(i)].load()) << "i=" << i;
```
luminousmining: same cast.hpp or define count at size_t

sources/resolver/cpu/tests/thread_pool.cpp
```diff
@@ -0,0 +1,131 @@
...
+    {
+        s.store(0);
+    }
+
+    pool.setCallback(
+        [&](uint64_t const lo, uint64_t const hi, uint32_t const)
+        {
+            for (uint64_t i{ lo }; i < hi; ++i)
+            {
+                seen[static_cast<size_t>(i)].fetch_add(1);
```
luminousmining: cast.hpp

sources/resolver/cpu/tests/thread_pool.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+
+// runAsync() must return before the batch is done; wait() then blocks until it is. After
+// wait() returns, the whole range must have been covered exactly once.
+TEST(CpuThreadPool, runAsyncThenWaitCoversAllIndices)
+{
+    constexpr uint64_t            count{ 5000ull };
+    resolver::cpu::CpuThreadPool  pool{ 4u, 0ull };
+    std::vector<std::atomic<int>> seen(static_cast<size_t>(count));
```
luminousmining: cast.hpp

sources/resolver/cpu/tests/thread_pool.cpp
```diff
@@ -0,0 +1,131 @@
...
+            {
+                s.store(0);
+            }
+
+            pool.setCallback(
+                [&](uint64_t const lo, uint64_t const hi, uint32_t const)
+                {
+                    for (uint64_t i{ lo }; i < hi; ++i)
+                    {
+                        seen[static_cast<size_t>(i)].fetch_add(1);
```
luminousmining: cast.hpp

sources/resolver/cpu/tests/thread_pool.cpp
```diff
@@ -0,0 +1,131 @@
...
+// [0, count) must still be visited exactly once regardless of how finely it is sliced.
+TEST(CpuThreadPool, smallGrainStillCoversAllIndicesExactlyOnce)
+{
+    for (uint32_t const workers : { 2u, 4u, 8u })
+    {
+        for (uint64_t const grain : { 1ull, 3ull, 7ull })
+        {
+            constexpr uint64_t            count{ 1000ull };
+            resolver::cpu::CpuThreadPool  pool{ workers, 0ull };
+            std::vector<std::atomic<int>> seen(static_cast<size_t>(count));
```
luminousmining: cast.hpp

sources/resolver/cpu/tests/thread_pool.cpp
```diff
@@ -0,0 +1,131 @@
...
+            {
+                s.store(0);
+            }
+
+            pool.setCallback(
+                [&](uint64_t const lo, uint64_t const hi, uint32_t const)
+                {
+                    for (uint64_t i{ lo }; i < hi; ++i)
+                    {
+                        seen[static_cast<size_t>(i)].fetch_add(1);
```
luminousmining: cast.hpp

sources/resolver/cpu/tests/thread_pool.cpp
```diff
@@ -0,0 +1,131 @@
...
+// run() must visit every index in [0, count) exactly once, across worker counts and
+// batch sizes including count < workers and count not divisible by workers.
+TEST(CpuThreadPool, runCoversAllIndices)
+{
+    for (uint32_t const workers : { 1u, 2u, 4u, 8u })
+    {
+        for (uint64_t const count : { 0ull, 1ull, 5ull, 1000ull })
+        {
+            resolver::cpu::CpuThreadPool  pool{ workers, 0ull };
+            std::vector<std::atomic<int>> seen(static_cast<size_t>(count));
```
luminousmining: cast.hpp

sources/algo/tests/bitwise.cpp
yuzi-co: Done in `8122a55` — `castU8`/`castU16` from `cast.hpp` instead of `static_cast`.

sources/common/cast.hpp
yuzi-co: Done in `8122a55` — removed the stray separator line.

sources/device/device.cpp
yuzi-co: Done in `8122a55` — `getMinimumKernelExecuted` now takes a `config` const-ref and returns `config.occupancy.kernelMinimunExecuteNeeded` directly.

sources/resolver/cpu/tests/thread_pool.cpp
yuzi-co: Done in `8122a55` — hoisted to `constexpr std::array<uint32_t, 4> workerCounts{ 1u, 2u, 4u, 8u };`.

sources/resolver/cpu/tests/thread_pool.cpp
yuzi-co: Done in `8122a55` — hoisted to `constexpr std::array<uint64_t, 4> counts{ 0ull, 1ull, 5ull, 1000ull };`.

sources/resolver/cpu/tests/thread_pool.cpp
yuzi-co: Done in `8122a55` — hoisted to `constexpr std::array<uint32_t, 3> workerCounts{ 2u, 4u, 8u };`.

sources/resolver/cpu/tests/thread_pool.cpp
yuzi-co: Done in `8122a55` — hoisted to `constexpr std::array<uint64_t, 3> grains{ 1ull, 3ull, 7ull };`.

sources/resolver/cpu/tests/thread_pool.cpp
yuzi-co: Done in `8122a55` — using `castSize()` from `cast.hpp`.

sources/resolver/cpu/tests/thread_pool.cpp
yuzi-co: Done in `8122a55` — using `castSize()` from `cast.hpp`.

sources/resolver/cpu/tests/thread_pool.cpp
yuzi-co: Done in `8122a55` — using `castSize()` from `cast.hpp`.

sources/resolver/cpu/tests/thread_pool.cpp
yuzi-co: Done in `8122a55` — using `castSize()` from `cast.hpp`.

sources/resolver/cpu/tests/thread_pool.cpp
yuzi-co: Done in `8122a55` — using `castSize()` from `cast.hpp`.

sources/resolver/cpu/tests/thread_pool.cpp
yuzi-co: Done in `8122a55` — using `castSize()` from `cast.hpp`.

sources/resolver/cpu/tests/thread_pool.cpp
yuzi-co: Done in `8122a55` — using `castSize()` from `cast.hpp`.

sources/resolver/cpu/tests/thread_pool.cpp
yuzi-co: Done in `8122a55` — using `castSize()` from `cast.hpp`.

sources/resolver/cpu/tests/thread_pool.cpp
yuzi-co: Done in `8122a55` — using `castSize()` from `cast.hpp`.

PR #174, opened 2026-06-11
feat(stats): keep the hashrate meter alive across jobs (--internal_accumulate_hash)

yuzi-co:
Reworks this PR to the `--internal_accumulate_hash` design we converged on in the thread, replacing the time-windowed EWMA meter. The EWMA implementation is preserved on `feat/hashrate-ewma-meter` in case we ever want to revisit smoothing separately.

## Problem

Slow / memory-hard kernels (Xelis, Octopus, Ergo…) complete far fewer than `--internal_kernel_count` launches per job window. The meter is count-gated and reset on every job update — `updateBatchNonce()` zeroes the stored hashrate — so the threshold is never reached, `getHashrate()` returns 0, and the `HASHRATE` board is hidden entirely even while valid shares land.

## Fix

New Kernel-section flag **`--internal_accumulate_hash`** (default **`true`**):

- **`true`** — job and constant updates no longer reset the hash count or the elapsed window; only a **memory rebuild** (e.g. DAG) resets. The meter accumulates a true `nonces / elapsed` average that spans jobs, so slow kernels publish a real number out of the box and read `0` only when the GPU is genuinely idle.
- **`false`** — restores the previous per-job reset behaviour exactly.

The work loop also opens the meter window once at start-up: a non-epoch algorithm's first job is a constant update (not a memory update), so under accumulation its clock would otherwise never start.

## Changes

- `config` / CLI: new `--internal_accumulate_hash` bool wired into `DeviceOccupancy` (default `true`).
- `device`: gate the epoch/period reset and the `updateBatchNonce` reset behind `(!accumulate || memory update)`; memory rebuilds always reset; open the meter window once before the loop.
- docs: document the flag in `PARAMETERS.md`.
- test: `statistical` unit tests for the accumulate-vs-reset invariant.

## Verification

**Unit tests** (`unit_test`, AMD cross-compile): 119/123 pass. The 4 failures are the pre-existing `*AmdTest.allDeviceFindNonce` progpow-family cases, unrelated to this change. The 3 new `StatisticalHashrate.*` tests pass.

**Live, RX 9070 XT vs Octopus** (`de.conflux.herominers.com:1170`), same binary, ~95 s each:

| `--internal_accumulate_hash` | HASHRATE board | Hashrate |
|---|---|---|
| `true` (default) | renders every 10 s | steady **24.30 MH/s** |
| `false` | never appears | 0 H/s (current behaviour, reproduced) |


sources/device/device.cpp
```diff
@@ -590,9 +590,15 @@ bool device::Device::updateJob()
     uint64_t const currentAtomicMemory{ synchronizer.memory.get() };
 
     ////////////////////////////////////////////////////////////////////////////
+    common::Config const& config{ common::Config::instance() };
     if (nextjobInfo.epoch != currentJobInfo.epoch || nextjobInfo.period != currentJobInfo.period)
     {
-        miningStats.reset();
+        // When accumulating, the meter window spans jobs; only a memory rebuild
```
luminousmining: remove comment

sources/device/device.cpp
```diff
@@ -636,13 +642,15 @@ bool device::Device::updateJob()
     }
 
     ////////////////////////////////////////////////////////////////////////////
-    updateBatchNonce();
+    // Reset the meter on memory rebuilds always; on plain job/constant updates
```
luminousmining: remove comment

sources/device/device.cpp
```diff
@@ -636,13 +642,15 @@ bool device::Device::updateJob()
...
+    // only when not accumulating, so the window can span jobs.
+    updateBatchNonce(false == config.occupancy.accumulateHash || true == needUpdateMemory);
```
luminousmining: move `false == config.occupancy.accumulateHash || true == needUpdateMemory` in variable `resetStats`

sources/device/device.cpp
```diff
@@ -656,8 +664,15 @@ void device::Device::updateBatchNonce()
 
     ////////////////////////////////////////////////////////////////////////////
     miningStats.setBatchNonce(resolver->getBlocks() * resolver->getThreads() * internalLoop);
-    miningStats.resetHashrate();
-    miningStats.reset();
+
+    ////////////////////////////////////////////////////////////////////////////
```
luminousmining, line 665: remve comment

sources/device/device.cpp
```diff
@@ -691,6 +706,12 @@ void device::Device::loopDoWork()
     ////////////////////////////////////////////////////////////////////////////
     computing.store(true, boost::memory_order::seq_cst);
 
+    ////////////////////////////////////////////////////////////////////////////
+    // Open the hashrate window once at start-up. When accumulating, a job that
```
luminousmining: remove comment

sources/statistical/tests/statistical.cpp
```diff
@@ -0,0 +1,82 @@
+#include "statistical/statistical.hpp"
+
+#include <chrono>
+#include <thread>
+
+#include <gtest/gtest.h>
+
+
+using statistical::Statistical;
```
luminousmining: never declare `using` for namespace out of scope.

sources/statistical/tests/statistical.cpp
```diff
@@ -0,0 +1,82 @@
+#include "statistical/statistical.hpp"
```
luminousmining: use `<>` instead of `""`
Respect ordre include

sources/statistical/tests/statistical.cpp
luminousmining, line 1: Function name -> `lowerCase`
Exemple: `WorkedWindowPublishesPositiveHashrate` -> `workedWindowPublishesPositiveHashrate`

